PR #202, opened 2026-03-29
fix: cancel button race condition with stream abort and force cancel

eous:
The cancel endpoint emitted a 'cancelled' SSE event before the worker thread terminated. The frontend transitioned to "send" mode prematurely, so the next send got rejected with "Already processing a request."

Backend:
- Providers expose SDK stream handle via cancel_ref parameter so cancel() can close the HTTP connection and unblock iteration
- `_CancelRef` list subclass eagerly updates `_cancel_stream` when the provider appends its SDK stream handle, and immediately closes the stream if cancel was already requested before the first chunk arrived — fixes the window where a slow-to-respond model could leave the worker stuck despite cancellation
- Generation counter prevents orphaned threads from mutating messages or clearing cancel state after force cancel
- `_check_cancelled()` added between retry attempts in `_try_stream`
- Server polls (async, non-blocking) for cancelled worker to exit
- Force cancel (force:true) abandons stuck worker, keeps cancel event set so subprocesses are killed, guards against spurious SSE events

Frontend:
- 'cancelled' shows "Cancelling..." then escalates to "Force Stop" after 2s for a harder cancel that abandons the worker immediately
- 10s safety timeout auto-recovers if stream_end never arrives
- busy_error re-enables stop button instead of showing send
- Timeout cleanup in disconnectSSE, stream_end, and force .then()
- Layout shift prevention (min-width, white-space: nowrap)
- aria-label updates for accessibility

Tests:
- 17 new tests: stream close, error suppression, cancel_ref population, transport error conversion, non-cancel exception propagation, retry cancellation check, `_CancelRef` eager population and close-on-cancel, force-cancel generation tracking and orphaned-thread protection

---

📱 Kick off Copilot coding agent tasks wherever you are with [GitHub Mobile](https://gh.io/cca-mobile-docs), available on iOS and Android.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -452,9 +452,12 @@ after `/clear` or `/new` commands).
 {"type": "clear_ui"}
 ```
 
-**`cancelled`** -- the generation was cancelled by the user (via the Stop
-button or `POST /v1/api/cancel`). The client should finalize any in-progress
-assistant message with whatever partial content was streamed.
+**`cancelled`** -- a cancel request was acknowledged (via the Stop button or
+`POST /v1/api/cancel`). This signals that cancellation is in progress, not
+that it is complete. The worker thread may still be finishing — wait for
+`stream_end` before transitioning to a ready state. The client should clear
+any in-progress assistant rendering but not re-enable the send button until
+`stream_end` arrives.
 
 ```json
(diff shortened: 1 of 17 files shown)


eous:
@copilot please review this pull request.

PR #207, opened 2026-03-29
refactor: add is_error to on_tool_result protocol, remove text heuris…

eous:
…tics

Add is_error keyword arg to SessionUI.on_tool_result() so tools report errors structurally. Server and JS client no longer guess from output text prefixes — each tool sets the flag at the source.

Bash tool: exit code >= 2 is error, exit code 1 is ambiguous (grep no-match). History reconstruction keeps text heuristic as fallback for pre-migration data.

Update SDKs (Python + TypeScript), test mocks, docs, and diagrams.


eous:
@copilot please review this pull request

PR #327, opened 2026-04-06
fix: preserve Gemini tool-call metadata for thought_signature round-trip

sillyWillieBilly:
## Summary

Gemini's OpenAI-compatible endpoint embeds provider-specific fields (e.g. `thought_signature`) inside the `function` object of each tool call. These fields **must** survive the tool-call → tool-result round-trip or subsequent API calls fail with a 400 error. This broke after dc3a1b7 because the metadata was being silently discarded during message serialization.

## Problem

When using Gemini 2.5+ models (Flash, Pro) through the OpenAI-compatible endpoint, any workstream that involves function calling fails with:

```
400 Please ensure that the function response turn comes immediately
after a function call turn, with all function call IDs responded to.
```

This happens because `thought_signature` (a Gemini-specific field nested inside `function` alongside `name` and `arguments`) is stripped during the tool-call accumulation in `session.py` and never restored when sending the next request.

### Screenshots

**Gemini Flash function call error:**
gemini-flash-error

**Gemini Pro function call error:**
gemini-pro-error

## Solution

A two-layer metadata capture protocol that preserves provider-specific extras at both the tool-call level and inside the `function` sub-dict:

### Core metadata protocol (`_openai_chat.py`)
- Added `_KNOWN_TC_KEYS` / `_KNOWN_FN_KEYS` frozensets to identify standard fields
- **Streaming path**: after extracting `id`, `name`, `arguments_delta`, calls `model_dump(exclude_none=True, exclude_unset=True)` and captures anything not in the known-key sets into `ToolCallDelta.extra_fields` (top-level extras) and `extra_fields["_fn_extra"]` (function-level extras like `thought_signature`)
- **Non-streaming path**: same capture logic, extras written directly to the tool-call dict

### Round-trip promotion (`_openai_common.py`)
- `sanitize_messages()` now promotes `_extra` keys back to each tool-call's top level, and `_fn_extra` keys back into the `function` sub-dict, before the message list is sent to the API

### Data model (`_protocol.py`)
- `ToolCallDelta`: new `extra_fields: dict` field (default `{}`)
- `ModelCapabilities`: new `preserve_tool_metadata: bool` flag

### Accumulator (`session.py`)
- Streaming accumulator merges `tcd.extra_fields` into `tc["_extra"]` using `getattr()` for backward compatibility with test fakes

### Additional fixes included in this PR

1. **`chat_template_kwargs` rejected by Google** — Two-layer guard:
   - `model_registry.py`: `_resolve_openai_provider()` detects `googleapis.com` in `base_url` → returns `"google"` provider instead of `"openai-compatible"`
   - `session.py`: runtime URL guard in `_provider_extra_params()` checks `self.client.base_url` for `googleapis.com` and returns `None`

2. **`cancel_on_approval` never wired** — `_judge_cfg()` was missing the `cancel_on_approval` setting, and `judge.py:_evaluate_single()` unconditionally cancelled in-flight LLM calls regardless of config:
   - `session.py`: added `cancel_on_approval=cs.get("judge.cancel_on_approval")` to `_judge_cfg()`
   - `judge.py`: guarded `future.cancel()` behind `self._config.cancel_on_approval`

## Impact on non-Gemini models

None. For standard OpenAI / local models, `model_dump(exclude_none=True, exclude_unset=True)` returns only known keys, so `extra_fields` stays empty. The accumulator's `if _extra:` check means nothing is stored or promoted. All existing behavior is preserved.

## Testing

- All **3468 tests** pass (`pytest -m "not live" -q`), 0 failures
- Verified end-to-end with Gemini 2.5 Flash and Pro — `thought_signature` survives the full round-trip
- Non-streaming and streaming paths both tested
- Backward compatible with `FakeToolDelta` in `test_cancel.py` (uses `getattr` fallback)

## Files changed

| File | Change |
|------|--------|
| `turnstone/core/providers/_openai_chat.py` | Two-layer metadata capture (streaming + non-streaming) |
| `turnstone/core/providers/_openai_common.py` | `sanitize_messages()` promotes `_extra` / `_fn_extra` back |
| `turnstone/core/providers/_protocol.py` | `extra_fields` on `ToolCallDelta`, `preserve_tool_metadata` on `ModelCapabilities` |
| `turnstone/core/providers/_google.py` | `preserve_tool_metadata=True` in `_GOOGLE_DEFAULT` |
| `turnstone/core/session.py` | Accumulator merges extras; URL guard; `cancel_on_approval` wiring |
| `turnstone/core/model_registry.py` | Google URL detection in `_resolve_openai_provider()` |
| `turnstone/core/judge.py` | Guard `cancel_event` behind `config.cancel_on_approval` |


eous:
dc3a1b7 Seems like just fronted code? I’ll take a look at the bug though.

PR #353, opened 2026-04-14
replace bitnami pgbouncer with edoburu pgbouncer

pizzaandcheese:
replaced bitnami pgbouncer with edoburu pgbouncer container and updated environment variables to fit

In August of 2025 Bitnami sunset their free pgbouncer containers, removing all available tags from the [repo](https://hub.docker.com/r/bitnami/pgbouncer/).

There is a [legacy repo](https://hub.docker.com/r/bitnamilegacy/pgbouncer) but it is not guaranteed to remain up forever and will not receive any further updates.

Therefore I propose the use of [edoburu's pgbouncer](https://hub.docker.com/r/edoburu/pgbouncer/)


eous:
Thanks for the PR — verified the claim and the replacement is correct:

- Bitnami's `pgbouncer` repo is indeed empty (0 tags on Docker Hub)
- edoburu's image is active, small (15MB Alpine), and all the env var names in this PR match their [README](https://github.com/edoburu/docker-pgbouncer)
- Adding `AUTH_TYPE: scram-sha-256` is a nice upgrade over edoburu's md5 default

**One bug though — port mismatch:**

Bitnami's pgbouncer listened on `6432` by default. edoburu's [entrypoint.sh](https://github.com/edoburu/docker-pgbouncer/blob/master/entrypoint.sh) sets `listen_port = ${LISTEN_PORT:-5432}`, so it listens on **5432** internally.

As written, the compose snippet keeps `ports: "6432:6432"` and the healthcheck / connection instructions all target `6432` — but nothing will be listening there. The container-to-container DNS lookup (`TURNSTONE_DB_URL=postgresql://...@pgbouncer:6432/...`) will fail too.

Fix is one line — add `LISTEN_PORT: "6432"` so it listens where the rest of the config expects:

```yaml
environment:
  DB_HOST: postgres
  DB_PORT: "5432"
  LISTEN_PORT: "6432"
  DB_NAME: ${POSTGRES_DB:-turnstone}
  ...
```

Happy to merge once that's added.

pizzaandcheese:
Sorry about merging upstream, I misclicked.

Do you think it would be worth changing the other instances of 6432 to fit what pgbouncer is publishing? i.e. changing the ports and healthcheck to 5432. This could make it more transparent for the user as they would only need to update their postgres yaml to stop publishing 5432 when adding pgbouncer so they don't conflict.

This way the postgres URL on the servers wont need to be updated with a different port.

If so we could add a note in the docs that mention to remove the ports from postgres if adding pgbouncer.

PR #355, opened 2026-04-15
Add Slack channel adapter with Socket Mode support

daoxley:
Adds a Slack channel adapter mirroring the Discord adapter pattern:

- Socket Mode connection
- Per-user session management in channels via configurable slash command
- SSE-based event consumption from server nodes
- Tool approval buttons with policy evaluation support
- DM routing without slash command
- Session recovery after restart via recoverable route keys

New files:
- turnstone/channels/slack/bot.py
- turnstone/channels/slack/config.py
- turnstone/channels/slack/__init__.py
- tests/test_channel_slack.py

Updated:
- turnstone/channels/cli.py — adds Slack CLI arguments
- pyproject.toml — adds slack extra and mypy overrides

Usage:
- Install with Slack support:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,8 @@ ddg = ["ddgs>=9.0"]
 discord = ["discord.py>=2.4"]
 tls = ["lacme>=1.0.5"]
 sandbox = ["sympy>=1.13", "numpy>=2.0", "scipy>=1.14", "pytest>=9.0"]
-all = ["turnstone[console,anthropic,postgres,discord,ddg,tls,sandbox]"]
+slack = ["slack-bolt>=1.18", "aiohttp>=3.9"]
+all = ["turnstone[console,anthropic,postgres,discord,ddg,tls,sandbox,slack]"]
 
 [project.scripts]
 turnstone = "turnstone.cli:main"
@@ -183,3 +184,8 @@ warn_unused_ignores = false
 [[tool.mypy.overrides]]
 module = "tests.*"
 disallow_untyped_defs = false
+
+[[tool.mypy.overrides]]
+module = ["slack_bolt", "slack_bolt.*", "slack_sdk", "slack_sdk.*"]
+ignore_missing_imports = true
+disallow_untyped_calls = false
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ Repository = "https://github.com/turnstonelabs/turnstone"
 Issues = "https://github.com/turnstonelabs/turnstone/issues"
 
 [project.optional-dependencies]
-test = ["pytest>=9.0", "pytest-cov>=6.0", "croniter>=3.0"]
+test = ["pytest>=9.0", "pytest-cov>=6.0", "croniter>=3.0", "slack-bolt>=1.18", "aiohttp>=3.9"]
 dev = ["ruff>=0.9", "mypy>=1.14"]
 console = ["croniter>=3.0"]
 anthropic = ["anthropic>=0.39"]
--- a/tests/test_channel_slack.py
+++ b/tests/test_channel_slack.py
@@ -146,7 +146,7 @@ class TestStreamingMessage:
         client = AsyncMock()
         client.chat_postMessage = AsyncMock(return_value={"ok": True, "ts": "123"})
         client.chat_update = AsyncMock(return_value={"ok": True})
-        
+
         sm = StreamingMessage(client=client, channel="C1", edit_interval=999.0)
 
         _run(sm.append("hello "))
--- a/tests/test_channel_slack.py
+++ b/tests/test_channel_slack.py
@@ -1,0 +1,552 @@
+"""Tests for the Slack channel adapter (bot, config, CLI)."""
+
+from __future__ import annotations
+
+import asyncio
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+slack_bolt = pytest.importorskip("slack_bolt")
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _run(coro):  # type: ignore[no-untyped-def]
+    """Run an async coroutine in a fresh event loop (no pytest-asyncio needed)."""
+    return asyncio.run(coro)
+
+
+def _make_slack_event(
+    *,
+    bot_id: str | None = None,
+    subtype: str | None = None,
+    channel: str = "C01SAPU5414",
+    channel_type: str = "channel",
+    thread_ts: str = "1234567890.000100",
+    user: str = "U12345",
+    text: str = "hello",
+    ts: str = "1234567890.000200",
+) -> dict[str, object]:
+    event: dict[str, object] = {
+        "channel": channel,
+        "channel_type": channel_type,
+        "user": user,
+        "text": text,
+        "ts": ts,
+        "thread_ts": thread_ts,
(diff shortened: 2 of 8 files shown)


eous (changes requested):
## Code Review — Slack Channel Adapter

Thanks for the PR — structure cleanly mirrors the Discord adapter and the dependency footprint is minimal. A few issues need addressing before merge.

---

### Critical (must fix)

**1. Unified startup path — Slack-only bypasses the channel HTTP server, and Slack is ignored when Discord is also set**
- `turnstone/channels/cli.py:233-266` and `turnstone/channels/cli.py:268-304`
- The Slack-only branch starts `TurnstoneSlackBot` directly and `return`s. That path never creates the channel HTTP app, never exposes `/v1/api/notify`, never registers the channel service, and never sends heartbeats. Additionally, `if args.slack_token and not args.discord_token:` means that when both tokens are provided, Slack is silently skipped and only Discord starts.
- Fix: collapse into one startup path that builds an `adapters: dict[str, ChannelAdapterLike]`, always creates the channel app, registers/heartbeats the channel service, and `asyncio.gather`s all adapter `.start()` coroutines with `server.serve()`. Sketch:
  ```python
  adapters: dict[str, ChannelAdapterLike] = {}
  if args.discord_token:
      adapters["discord"] = TurnstoneBot(...)
  if args.slack_token:
      adapters["slack"] = TurnstoneSlackBot(...)
  channel_app = create_channel_app(adapters, storage, jwt_secret=jwt_secret)
  await asyncio.gather(*(a.start() for a in adapters.values()), server.serve())
  ```

**2. Notification reply routing is internally inconsistent and currently won't work**
- `turnstone/channels/slack/bot.py:380-382`, `turnstone/channels/slack/bot.py:387-395`, `turnstone/channels/slack/bot.py:872-901`, `turnstone/channels/_http.py:136-141`
- Three concrete defects:
  - `_on_message` returns to `_handle_dm` at line 381 *before* the `_notify_ws_map` check at 387 — DM notification replies never hit the notification-reply path.
  - `send()` parses a 2-part `channel_id` as `channel:thread_ts` (line 875), but `send_notification()` parses the same 2-part form as `channel:user_id` (line 894). Same shape, contradictory meanings.
  - `_handle_notify` in `_http.py:136-141` passes whatever `channel_id` the API supplies through as-is; a bare single-segment channel ID will never trigger `_track_notification` (which requires `target_user_id` from parts[1]).
- Additionally, unlike Discord there's no `_notify_reply_channels`-style mechanism to route the assistant's next response back into the notification conversation.
- Fix: define one canonical notification-target shape and use it consistently on send *and* reply. Consider an explicit dataclass instead of overloading `channel_id`:
  ```python
  @dataclass(frozen=True)
  class SlackRoute:
      channel: str
      user_id: str | None = None
      thread_ts: str | None = None
  ```
  Resolve notification replies before generic DM routing (or inside `_handle_dm`), and add a response-forwarding mechanism similar to Discord's DM reply flow. See `turnstone/channels/discord/cog.py:230-281` and `turnstone/channels/discord/bot.py:800-815` for the reference shape.

**3. Plan review events are non-actionable in Slack**
- `turnstone/channels/slack/bot.py:755-760`, `turnstone/channels/_routing.py:352-358`
- `PlanReviewEvent` only posts a text message. There is no button/modal that calls `router.send_plan_feedback(...)`, so any workstream that pauses for plan review dead-ends for Slack users.
- Fix: mirror the Discord interactive flow — track pending plan-review state per `ws_id`, post approve/request-changes actions, and wire the handlers to `router.send_plan_feedback(ws_id, ..., feedback)`.

**4. Approval buttons lack per-user authentication**
- `turnstone/channels/slack/bot.py:444-486`, `turnstone/channels/slack/bot.py:791-844`
- `_on_approve` / `_on_deny` read `body["actions"][0]["value"]` (which is `f"{ws_id}|{correlation_id}"`) and call `router.send_approval` without verifying that `body["user"]["id"]` is the session owner. In a channel where an approval card is posted, any member who can see the message can approve or deny it — including dangerous tool calls initiated by someone else. Unlike Discord where the approval is scoped to a user-owned thread, Slack approvals in a shared channel are visible to everyone.
- Fix: record the session owner's Slack user ID alongside the pending approval TS (e.g. extend `_pending_approval_ts` or add a parallel `_pending_approval_owner` map), and in `_on_approve` / `_on_deny` reject with an ephemeral message when `body["user"]["id"]` doesn't match:
  ```python
  owner = self._pending_approval_owner.get(ws_id)
  if owner and body["user"]["id"] != owner:
      await self._client.chat_postEphemeral(
          channel=body["container"]["channel_id"],
          user=body["user"]["id"],
          text="Only the session owner can approve this tool call.",
      )
      return
  ```

---

### Suggestions

- **Escape and size-cap approval previews before interpolating into `mrkdwn` blocks** (`turnstone/channels/slack/bot.py:799-840`). Triple-backtick fences don't neutralize `<@userid>` mentions and can be broken by backticks in the preview. Mirror the escaping/truncation the Discord adapter performs.
- **Fail fast if `--slack-token` is set without `--slack-app-token`** (`turnstone/channels/cli.py:221-266`). Slack is currently considered "configured" once `bot_token` is present; without `app_token`, Socket Mode dies at startup. Validate both up front.

### Minor

- `turnstone/channels/cli.py:224-229` — error text still says only "`Set --discord-token or $TURNSTONE_DISCORD_TOKEN.`"; include the Slack env var.
- `turnstone/channels/slack/bot.py:429-434` — pass `client_type="chat"` to `get_or_create_workstream` for parity with Discord (Slack sessions currently don't get the chat-specific system prompt).
- `turnstone/channels/slack/bot.py:99, 139, 338, 464, 486, 753, 775` — `except Exception: log.debug(...)` without `exc_info=True`. Match Discord's pattern or use `log.exception`.
- `turnstone/channels/slack/__init__.py` is empty — add at least a module docstring.
- `pyproject.toml:56` — `aiohttp>=3.9` is already a transitive of slack-bolt; drop the explicit pin to avoid drift.

---

### Testing

The suite covers config defaults, streaming, basic `_on_message` filtering, session isolation, and simple `_on_ws_event` dispatch. Notable gaps:

- No test for approval-button rejection when a non-owner clicks (gates Critical #4).
- No Slack-only CLI test that `/v1/api/notify` is exposed, and no Discord+Slack combined startup test (gates Critical #1).
- No notification delivery + reply round-trip test (gates Critical #2).
- No plan-review interaction test (gates Critical #3).
- DM session continuity across multiple messages not exercised.

**Also: the Slack test suite is silently skipped in default CI.** `pyproject.toml:46` declares `test = [...]` without `slack-bolt`, CI runs `pip install -e ".[test]"` (`.github/workflows/ci.yml:46`), and `tests/test_channel_slack.py` starts with `pytest.importorskip("slack_bolt")`. Either add `slack-bolt` to the `test` extra, or extend the CI install to `.[test,slack]`, otherwise this suite provides zero coverage in CI.

---

### Positive

- Token factories (`_server_token_factory` / `_console_token_factory`) are wired correctly for rotation.
- `_MAX_NOTIFY_TRACKING=100` FIFO eviction is correctly ported.
- `chunk_message` is reused rather than reimplemented.
- Mypy overrides are scoped to `slack_bolt` / `slack_sdk` only — the rest of the project keeps strict typing.
- Socket Mode choice avoids the signing-secret / public-URL burden for the common case.
- Config docstring walks operators through Slack app creation step-by-step.

eous:
Thank you

PR #363, opened 2026-04-16
Feat/attachment followups

eous:


eous:
Pushed 2ad95d2 addressing the Copilot feedback. Summary:

**Real correctness bug — fixed (5 threads):** `reserved_at` column added (migration 038), set on `reserve_attachments` and cleared on `mark_attachments_consumed` / `unreserve_attachments`. Sweep now scopes by reservation age rather than upload age, so the race that prompted the feedback (uploaded long ago + reserved fresh → swept mid-send) is gone. Backed by a partial index on `(reserved_at) WHERE reserved_at IS NOT NULL`. Threshold tightened from 4h → 1h since it now means "longest realistic single send". New regression test asserts the previously-broken case is preserved.

**False positives — declining (2 threads):**
- `_protocol.py:165` (`...` vs `pass`): every other Protocol method body in this file uses `...`, which is the canonical Python convention for Protocol stubs ([PEP 544](https://peps.python.org/pep-0544/) examples use it). Changing one method to `pass` would be inconsistent for no real benefit.
- `server.py:3850` (`await _orphan_sweep_task`): an `await` expression IS the side-effecting statement we want — it waits for the task to finish on shutdown. Linter false positive.

PR #445, opened 2026-04-28
feat(node): auto-detect node capabilities via kernel interfaces

eous:
## Summary

Closes the operator-burden gap the coordinator-harness shakedown surfaced: the `list_nodes` capability/region/role filtering surface had nothing to filter on because operators were launching nodes with empty metadata. Auto-detection now runs at server startup and populates `node_metadata` with sensible defaults — operators can still override any auto-detected key via the `[metadata]` section of `config.toml` (operator config writes win on the per-key upsert).

### What's detected — all kernel interfaces, no userspace binaries

| Key | Source |
|---|---|
| `gpu_count`, `gpu_vendor`, `gpu_vendors`, `gpus` | `/sys/class/drm/cardN/device/{vendor,device}` (PCI ID decode → NVIDIA / AMD / Intel / Apple) |
| `memory_gb` | `/proc/meminfo` (rounded down) |
| `cpu_model` | `/proc/cpuinfo` first `model name` line |
| `cloud_provider` | `/sys/class/dmi/id/{sys_vendor,bios_vendor,board_vendor,chassis_asset_tag,product_name}` |
| `cloud_region`, `cloud_zone`, `cloud_instance_type`, `cloud_instance_id` | IMDS probe — only fired AFTER DMI confirmed the cloud provider |

No `nvidia-smi` / `rocm-smi` / `lspci` subprocess — kernel interfaces are vendor-agnostic and don't depend on which optional package the operator happened to install.

### Hardening

The IMDS path is the security-relevant surface — its responses flow into `node_metadata` and from there into coord-LLM `list_nodes` context. Five rounds of defense-in-depth:

1. **Link-local IP literal for every cloud** — AWS, GCP, and Azure all probe `169.254.169.254` directly. GCP no longer resolves `metadata.google.internal`, eliminating the DNS-spoof channel.
2. **Body cap** — `resp.read(64 KiB)` on every IMDS read.
3. **Per-field cap + control-char scrub** — each persisted `cloud_*` string is trimmed to 256 chars and stripped of control characters before storage.
4. **`isinstance(doc, dict)` guards** — a non-conformant JSON body (list / scalar / null) returns clean `{}` instead of raising `AttributeError`.
5. **DMI gates IMDS** — a non-cloud host pays zero startup latency on the cloud path. Operator opt-out via `TURNSTONE_AUTO_CLOUD_METADATA=0` skips IMDS entirely; DMI-derived `cloud_provider` still surfaces because that's a kernel interface.

### Performance

- `collect_node_info()` runs in `asyncio.to_thread` from the server's lifespan handler — IMDS probe latency never blocks the event loop.
- GCP fans the three zone/machine-type/id probes concurrently via `ThreadPoolExecutor` so a misidentified host's worst case is one timeout window (~1 s) instead of three (~3 s).

### Code review

Ran the multi-stage review pipeline (4 finders + verify + dedupe) before this PR; all 9 confirmed findings landed in this commit:

- bug-1 (heterogeneous-GPU flat key)
- bug-2 (AWS isinstance guard)
- bug-3 (Azure isinstance guard)
- sec-1 (GCP IP literal)
- sec-2 (body + field caps)
- perf-1 (asyncio.to_thread)
- perf-2 (concurrent GCP probes)
- q-1 (`_read_text` docstring)
- q-3 (test class acronym capitalization)

## Test plan
- [x] `pytest -m "not live"` — 4807 pass (+11 net since baseline; 73 in test_node_info.py)
- [x] `ruff check` clean on every modified file
- [x] `mypy` clean on every modified file
- [ ] Manual: launch a node on a GPU host (any vendor) and verify `list_nodes` surfaces `gpu_count` + `gpu_vendor`
- [ ] Manual: launch on AWS/GCP/Azure and verify `cloud_provider` + region populate
- [ ] Manual: launch on baremetal Linux and verify zero IMDS-probe latency at startup
- [ ] Manual: set `[metadata] gpu_vendor = "operator-override"` in config.toml and verify the operator value wins over the auto-detected one


eous:
@copilot+gpt-5.4 Please perform a code review of this pull request.

PR #447, opened 2026-04-29
feat(coord): inline tool-batch construct replaces approval dock

eous:
## Summary

UX polish for the coordinator chat: replace the pinned bottom approval-dock with one inline chat-stream construct per dispatch turn that pairs each tool call with its result and embeds the approval gate.

- `.coord-tool-batch--solo` for a single serial call; `.coord-tool-batch--parallel` for ≥2 calls (rows share a left rail + per-row tick so they read as siblings of one assistant decision).
- Lifecycle inside one construct: tool call → optional judge verdict / "judge evaluating…" placeholder → optional Approve/Deny/Always action row → tool_result paired under the originating row → resolved status pill (✓ approved / ✗ denied / ✓ auto / ✗ error).

## Why

The dock didn't scale. A 10-call `spawn_workstream` fan-out filled the whole pane with repeated verdict chips, and the call → approval → result lifecycle lived on three disconnected surfaces (inline `.msg.tool` bubble + dock + post-resolution `.msg.tool` result).

## What changed

- New tool-batch construct (`.coord-tool-batch` + rows + verdict line + result block + actions / status pill) in `coordinator.css` and DOM helpers in `coordinator.js`.
- SSE handlers (`approve_request`, `tool_info`, `intent_verdict`, `tool_result`, `approval_resolved`) routed through the construct.
- History replay aggregates each `assistant.tool_calls` array into one batch; orphan turns (no matching tool results = pending at reload) render as pending so SSE can upgrade in place.
- Pinned dock + dead helpers removed (`showApproval`'s old body, `hideApproval`, `claimApprovalFocus`, `applyJudgeVerdictToRow`, `applyJudgePendingToRow`, `ensureDctxAfterRow`, `removeRationale`, `setApprovalButtonsDisabled`, `appendToolCall` wrapper, `window.coordApprove`). ~248 lines of dead `.approval-dock` CSS gone too.

## Critical fix bundled

Page reload during a pending approval used to lock the operator out of acting in the new tab — history replay registered the call_ids as resolved-approved before SSE connected, and the SSE replay then folded into the existing resolved batch without painting actions. Now `init()` pre-scans every tool message's `tool_call_id`, marks the assistant turn pending if any call_id has no matching tool result, and `appendToolBatch` upgrades a non-pending existing batch in place when SSE arrives.

## Polish from review

- `_appendVerdictLineTo` dedupes via `row.dataset.verdictSig` so reconnect storms don't rebuild unchanged verdict lines.
- `judgeVerdicts` Map soft-capped at 500 entries (FIFO eviction).
- `toolRows` entries hold `{batch, row}` only.
- `_scheduleScroll` coalesces `scrollTop` writes through `requestAnimationFrame`.
- Rationale `<details>` now inserts after the verdict line (was tail-appending, breaking ordering once a result landed).
- `.coord-tool-batch--error` wired (lifted from row errors at construction + tool_result time).
- `approval_resolved` falls back to DOM lookup when `activeBatch` is null (cross-tab safety).
- `_buildStatusPill` extracted; reused by `_morphBatchResolved` and `appendToolBatch`'s resolved-replay branch.

## Test plan

- [ ] Live: 10× parallel `spawn_workstream` fan-out — verify batch frame, per-row verdicts, single Approve resolves all, results pair under their rows.
- [ ] Live: serial single-tool turn — verify solo construct, no rail.
- [ ] Live: tool with auto-approval — verify ✓ auto pill, no action row.
- [ ] Live: deny path — verify ✗ denied pill, rows dim.
- [ ] Live: error path — verify `--error` modifier on the batch + ✗ error result line.
- [ ] Reload mid-pending approval — verify operator can still approve in the new tab.
- [ ] Reload after approval resolved — verify resolved status pill renders, results paired.
- [ ] SSE reconnect during pending — verify no double-render, verdict line not torn down.
- [ ] Cross-tab resolution — verify the second tab morphs to resolved when the first tab approves.
- [ ] Mobile (<700px) — verify action row stacks, hit targets ≥44px.
- [ ] Children-tree approval block (right sidebar) untouched — still works.

### DIFF
--- a/tests/test_coordinator_page.py
+++ b/tests/test_coordinator_page.py
@@ -55,13 +55,16 @@ def test_uppercase_hex_rejected(client):
 
 
 def test_coordinator_js_exposes_inline_approval_helpers():
-    """Smoke guard for the Chunk 3 frontend wiring — the new helper
-    function names must remain reachable in the served JS so a refactor
-    accidentally renaming/removing them surfaces here instead of in
-    production where the children-tree's inline approve/deny buttons
-    silently stop rendering. Asserts string presence only — no DOM
-    parsing — since coord.js has no JS test framework today (per the
-    plan's testing notes)."""
+    """Smoke guard for two layers of the coord chat frontend: the
+    children-tree inline approve/deny block (the original Chunk 3
+    landing) and the PR #447 tool-batch construct that replaced the
+    pinned approval dock for the coord-self surface.  Both layers'
+    helper symbols must remain reachable in the served JS so a
+    refactor that accidentally renames or removes them surfaces here
+    instead of in production where the affected gates silently stop
+    rendering.  Asserts string presence only — no DOM parsing —
+    since coord.js has no JS test framework today (per the plan's
+    testing notes)."""
     from pathlib import Path
 
     coord_js = Path(__file__).resolve().parent.parent / (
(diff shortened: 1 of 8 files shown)


eous:
@copilot+gpt-5.4 Please perform a code review.

PR #450, opened 2026-04-29
fix(server): accept --skip-permissions CLI flag

OriginalOrangeXD:
## Summary

The server's `--help` epilog (`turnstone/server.py:3501`) and `compose.yaml` (lines 81, 204) both reference `--skip-permissions`, but the argparser never defined it. Any container started with `SKIP_PERMISSIONS=1` exited immediately with:

```
turnstone-server: error: unrecognized arguments: --skip-permissions
```

## Change

Wires the flag through to `app.state.skip_permissions`, OR-ing it with the existing `tools.skip_permissions` config-store setting so the stored value still works on its own. CLI-first ordering matches the prevailing idiom in `main()` (e.g. `args.api_key or os.environ.get(...)`, `mcp_config_cli or config_store.get(...)`).

One file, ~7 added / 2 modified lines. No schema, migration, API, or wire-format changes.

## Test plan

- [x] Reproduced the original `unrecognized arguments` crash with `SKIP_PERMISSIONS=1 docker compose --profile production up`
- [x] Rebuilt image with this patch — server accepts the flag, no crash
- [ ] Verify `--skip-permissions` alone (without env var) still works
- [ ] Verify `tools.skip_permissions=true` in config store still works without the flag

### DIFF
--- a/turnstone/server.py
+++ b/turnstone/server.py
@@ -3544,6 +3544,11 @@ def main() -> None:
         default=8080,
         help="Port to listen on (default: 8080)",
     )
+    parser.add_argument(
+        "--skip-permissions",
+        action="store_true",
+        help="Auto-approve all tool calls (no confirmation prompts)",
+    )
     # MCP config path is bootstrap-critical (needed before ConfigStore for tool loading)
     parser.add_argument(
         "--mcp-config",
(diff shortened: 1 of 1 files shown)


eous:
Are you only running server or console + server?
Screenshot From 2026-04-29 13-55-56
Since I don't have any telemetry I am not sure how many turnstone-server only people are out there. If you are turnstone-server only I see the need for this flag, but if you are running console the intended way to do this is by creating approval policies through the admin ux.

PR #455, opened 2026-04-30
fix(metacog): N>=3 streak detector + side-channel reminders

eous:
## Summary

Two-commit cleanup of the metacognitive nudge stack — fixes a stuck-loop detection regression and the system-reminder UI display bug surfaced by `1.5.0aN` testing.

### Commit 1 — `fix(metacog): N>=3 streak detector + drop redundant error-prefix list` (8c68ee9)

Cleanup pass on the per-batch advisory hook in `_run_loop`:

- `RepeatDetector` switched to consecutive-streak semantics (different sig resets count). Threshold bumped 2→3 to stop firing on legit transient retries.
- Deny path routed through `_report_tool_result(is_error=True)` so `_tool_error_flags` becomes the **single is_error source of truth**. The error-prefix string list (`"Error" / "JSON parse error" / …`) is gone.
- Errored tool calls now count toward the repeat streak again (the metacog split unintentionally introduced a "skip errors" branch that masked the stuck-loop pattern the warning was meant to catch).
- `metacognition._COOLDOWN_SECS` aligned to 300s (matches `MemoryConfig.nudge_cooldown` and the config-store default).
- Per-batch advisory block extracted to `ChatSession._apply_post_execute_advisories` for testability.
- Frontend: `appendToolOutput` now skips render when the parent approval block is denied, mirroring the history-replay guard. Necessary because the deny-path `is_error` change emits a `tool_result` SSE event that would otherwise duplicate the `✗ denied` badge.

### Commit 2 — `fix(session): metacog reminders ride a side-channel, not user content` (6c5bc73)

Architectural fix for the system-reminder UI display bug. User-channel nudges (`correction`, `denial`, `resume`, `start`, `completion`) were being spliced into `user_msg["content"]` permanently, leaking the `<system-reminder>` envelope into every consumer of `self.messages` — UI replay (band-aid'd by a regex strip), compaction, title generation, and any future channel adapter.

- `_attach_pending_user_reminders` (renamed from `_splice_pending_user_advisories`) writes the rendered reminders to `user_msg["_reminders"]` — a sibling key on the leading-underscore convention shared with `_attachments_meta` / `_provider_content`.
- New `_apply_reminders_for_provider` builds a transient shallow-copy at the provider boundary with the reminder spliced into `content`. Original message stays clean.
- `_mark_reminders_delivered` flips a `_reminders_delivered` flag after stream success so the model sees each reminder exactly once (the turn it advised). `_build_history` ignores the flag — UI replay parity preserved.
- `SessionUIBase.on_user_reminder` enqueues a `user_reminder` SSE event so other tabs render the bubble in lockstep.
- Frontend: new `.msg.user-reminder` bubble (yellow accent, pill-styled) anchored above the user message it advises, both live and on history replay.
- Pre-existing bug fix surfaced by the audit: cancel handlers cleared `_pending_tool_advisories` but not the user-channel buffer. Both now drain via a shared `_drain_pending_advisories` helper.
- `/history` regex strip removed — the side-channel approach makes it redundant.

## Test plan

- [x] `ruff check` — clean across 9 changed files
- [x] `mypy` — clean across 4 production modules
- [x] `tests/test_metacognition.py` — 86 pass (8 new `RepeatDetector` tests)
- [x] `tests/test_session.py` — 295 pass (`TestApplyRemindersForProvider`, `TestMarkRemindersDelivered`, `TestUpdateTokenTableMsgsParam`, `TestUserAdvisoryCancelClear`, `TestReminderSidechannelIsolation`, `TestSessionUIBaseUserReminderHook`, `TestApplyPostExecuteAdvisories`, plus rewritten splice/attach tests)
- [x] `tests/test_workstream_endpoints.py` — `TestBuildHistoryReminderPropagation` (presence/absence/empty/multi/coexist-with-attachments/malformed/all-malformed)
- [x] `tests/test_providers.py` — `test_sanitize_messages_strips_underscore_sibling_keys` covers `_reminders` + `_reminders_delivered`
- [x] **Full non-live suite: 4924 pass, 3 deselected, 127s**
- [ ] Manual smoke: open the web UI on a workstream, trigger a correction nudge, confirm the yellow reminder bubble renders above the user message both live and on `/history` reload
- [ ] Manual smoke: trigger 3 identical bash failures, confirm the streak warning fires once on the 3rd call and again every 3rd identical call after that

## Out of scope (follow-ups noted in code comments)

- Multi-tab parity for the `user_reminder` SSE event — no `user_message` event is emitted today, so a non-originating tab receives the reminder without a paired user-message render and may anchor to a stale prior bubble until `/history` reload corrects it. Documented as a known limitation in `app.js`'s SSE handler.
- Producer extraction (the `_queue_*` / `_attach_*` / `_splice_*` / `_apply_*` family of methods on `ChatSession`) — agreed to keep on the session for now and lift to a dedicated module in a follow-up PR.


eous:
Retargeting to `main` — this work is for the active development line, not a stable backport. Reopening fresh against `main` shortly.

PR #466, opened 2026-05-03
feat(coord): expose healthy model aliases per node on list_nodes

eous:
## Summary

- `list_nodes` rows now carry a top-level `model_aliases` field — the healthy aliases each node will accept on `spawn_workstream(model=...)` / `spawn_batch`. No HTTP fan-out: each server projects its registry into a `models` entry on `node_metadata` (`{alias, provider, healthy}` per alias) at lifespan startup, every 30s heartbeat tick, and immediately after `internal_model_reload`.
- New `turnstone_node_models_publish_total{outcome="written|skipped"}` Prometheus counter exposes the publish-cache hit rate so operators can graph backend health flapping or reload churn (`rate(skipped) / rate(total)` → ~1.0 in steady state).
- The provider-side model identifier (`cfg.model`) is intentionally dropped from the coord-visible payload — coordinators kept reaching for it when they should pass the local alias.

## Implementation notes

- `_collect_node_models_metadata` projects the registry; `_publish_models_metadata` is the cached writer used by both the heartbeat loop and the model-reload endpoint.
- Heartbeat shutdown awaits the cancelled task before deleting the auto-source rows, so an in-flight write can't resurrect the deleted row.
- Lifespan startup wraps the projection in its own try/except so a failure to project models doesn't take out the unrelated auto/config metadata write.
- Coord client tolerates missing/malformed `metadata.models` payloads (older nodes, mid-startup) and falls back to an empty `model_aliases` list.

## Test plan

- [x] `pytest tests/test_server_node_models_metadata.py tests/test_coordinator_client.py -m "not live"` — 12 + 121 green
- [x] `pytest -k "metric or models or coordinator_client or coordinator_tools or node_info or model_registry" -m "not live"` — 552 green
- [x] `ruff check` + `mypy` clean on all touched files
- [x] Multi-stage `/review` pipeline — all 11 findings landed (1 critical shutdown race, 4 minor, 3 nit, plus tests for shutdown race + model-reload integration)
- [ ] Manual smoke: bring a single-node cluster up, run `curl /metrics` and confirm `turnstone_node_models_publish_total{outcome="skipped"}` increments on each heartbeat after the first
- [ ] Manual smoke: in a coord session, call `list_nodes()` and confirm `model_aliases` matches the node's `/v1/api/_internal/model-status` aliases


eous:
Re: line 3522 `await _heartbeat_task` — CodeQL flagged this as a statement with no effect, but the `await` is the entire point of the shutdown-race fix. Cancelling a task with `.cancel()` schedules cancellation; `await` is what suspends the lifespan coroutine until the cancelled task actually finishes (including any in-flight `set_node_metadata_bulk` already inside `asyncio.to_thread`). Without it, the subsequent `delete_node_metadata_by_source(..., "auto")` could race the in-flight write and leave a resurrected `models` row. Treating this finding as a false positive.

PR #469, opened 2026-05-03
feat(providers): api_surface toggle + mistral medium reasoning fix

eous:
## Summary

- Adds an `api_surface` toggle (`chat` | `responses`) to `server_compat` so an `openai-compatible` model alias can be routed through either the Chat Completions or the Responses API. `openai-compatible` keeps Chat Completions as the default; operators flip individual aliases to Responses for endpoints that support it (Mistral medium on vLLM, Mistral cloud, future local backends).
- Adds a `vllm-mistral-medium` profile that pre-fills `api_surface=responses` on Detect so reasoning effort is delivered via `reasoning.effort` rather than `chat_template_kwargs`.
- **Drops the unconditional `reasoning_effort` injection into `chat_template_kwargs` for every openai-compatible request.** That hack only ever fit gpt-oss-style local templates; for any other backend (Mistral medium, Mistral cloud, Groq, OpenRouter) it could corrupt the rendered prompt. Operators who still want it now opt in via `server_compat.extra_body.chat_template_kwargs`.
- Fixes a soft-bug in `_run_agent` where the no-registry-route fallback path passed `alias=None` to capability/server_compat resolution, silently dropping per-model caps on the agent path.

## Why this PR exists

Hit while testing Mistral medium open-weights on vLLM: vLLM accepted the request but the model produced shakedown-style hallucinations because `chat_template_kwargs={"reasoning_effort": "high"}` was being passed to a chat template that didn't consume it. Going forward, any new model that needs Responses-API-native reasoning can be enabled with one click in the admin UI — no profile change needed.

## Notable details

- **Provider dispatch**: `create_provider("openai-compatible", api_surface="responses")` returns the `OpenAIResponsesProvider` singleton, reused across cloud OpenAI and openai-compat-with-Responses. The Responses provider's `provider_name` reports `"openai"` even when serving an openai-compat config — documented in the factory docstring; the only existing gate (`session.py:_provider_extra_params`) already allows both strings.
- **Cache invalidation**: `ModelRegistry.reload()` now evicts the cached provider singleton when `server_compat.api_surface` changes (not just when `cfg.provider` changes). Extracted `_api_surface_of(cfg)` helper so the lookup and eviction sites can't drift.
- **Server-side validation**: `admin_create_model_definition` / `admin_update_model_definition` reject unknown `api_surface` values at write time so an admin bypassing the `<select>` via direct API call can't poison a model alias.
- **Frontend allowlist**: detect's `suggested_server_compat.api_surface` is filtered against a closed set in `admin.js` so a hostile detect response can't smuggle a non-listed value into the form.

## Test plan

- [x] `ruff check` clean
- [x] `mypy` clean
- [x] `pytest -m "not live"` — **5117 passed** (one new regression test for `_run_agent` fallback alias)
- [x] New tests:
  - `test_create_provider_compat_responses_surface` — openai-compatible + responses returns Responses provider
  - `test_create_provider_invalid_api_surface` — unknown surface raises
  - `test_vllm_mistral_medium_responses_surface` — profile suggestion includes api_surface
  - `test_session_fallback_inherits_primary_alias_for_caps` — locks in the `_run_agent` fix
  - `test_operator_chat_template_kwargs_pass_through` — operator opt-in path for gpt-oss
- [ ] Manual: configure a Mistral medium vLLM endpoint, verify the admin Detect button suggests Responses API and a real chat completes with effort=high
- [ ] Manual: edit api_surface on a live alias and confirm the next request hits the new surface (i.e. cache eviction works)

## Migration note

If you operate a gpt-oss-style local model that consumed `reasoning_effort` from the chat template, edit that model's "Extra body params" in the admin UI to include:

```json
{"chat_template_kwargs": {"reasoning_effort": "high"}}
```

For every other openai-compatible backend the prior behaviour was a no-op at best and a prompt-corruption hazard at worst — no action needed.

### DIFF
--- a/tests/test_admin_model_registry_refresh.py
+++ b/tests/test_admin_model_registry_refresh.py
@@ -352,6 +352,90 @@ def test_update_endpoint_skips_refresh_on_empty_body(
     assert calls == []  # gate held: empty body did not trigger a refresh
 
 
+def test_create_rejects_invalid_api_surface(storage: SQLiteBackend) -> None:
+    """POST with a bogus server_compat.api_surface returns 400 rather than
+    persisting a value that would make get_provider() raise on every later
+    ChatSession init for the alias."""
+    _seed_model_def(storage, definition_id="m1", alias="local", model="m")
+    registry = _make_registry(alias="local", model="m")
+    client = _make_client(storage, registry)
+
+    resp = client.post(
+        "/v1/api/admin/model-definitions",
+        json={
+            "alias": "bad",
+            "model": "x",
+            "provider": "openai-compatible",
+            "base_url": "http://localhost:9000/v1",
+            "api_key": "sk-x",
+            "capabilities": {"server_compat": {"api_surface": "BOGUS"}},
+        },
+    )
+    assert resp.status_code == 400, resp.text
+    assert "api_surface" in resp.json()["error"]
+    # And the alias is not persisted
+    assert not registry.has_alias("bad")
+
+
+def test_create_rejects_non_canonical_api_surface(storage: SQLiteBackend) -> None:
+    """Strict validation: ' Responses ' / 'CHAT' don't round-trip through the
+    admin <select>, so they're rejected even though they'd survive a
+    case-insensitive membership check."""
+    _seed_model_def(storage, definition_id="m1", alias="local", model="m")
+    registry = _make_registry(alias="local", model="m")
+    client = _make_client(storage, registry)
+
+    for bad in (" responses ", "RESPONSES", "Chat"):
+        resp = client.post(
+            "/v1/api/admin/model-definitions",
+            json={
+                "alias": "noncanon",
+                "model": "x",
+                "provider": "openai-compatible",
+                "base_url": "http://localhost:9000/v1",
+                "api_key": "sk-x",
+                "capabilities": {"server_compat": {"api_surface": bad}},
+            },
+        )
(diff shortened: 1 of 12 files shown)


eous:
vllm config i landed on for mistral medium 3.5 — tensor-parallel 2, fp8 kv-cache, mistral tokenizer/config/load mode, tool-call + reasoning parsers wired up, and `default-chat-template-kwargs` pinning `reasoning_effort=high` so the engine handles effort instead of turnstone trying to.

```bash
docker run -d --gpus all -p 8000:8000 --ipc=host --ulimit memlock=-1 --ulimit stack=67108864 -v /mnt/models:/models --entrypoint bash vllm/vllm-openai:nightly -c 'apt update; apt install -y git; pip install git+https://github.com/mistralai/mistral-common.git git+https://github.com/huggingface/transformers.git && exec python3 -m vllm.entrypoints.openai.api_server --model /models/mistral-medium-3.5 --tensor-parallel-size 2 --tool-call-parser mistral --enable-auto-tool-choice --reasoning-parser mistral --tokenizer-mode mistral --config-format mistral --load-format mistral --served-model-name mistral-medium --gpu_memory_utilization 0.91 --kv-cache-dtype fp8 --max-num-seqs 1 --max-num-batched-tokens 4096 --default-chat-template-kwargs '"'"'{"reasoning_effort":"high"}'"'"' --override-generation-config '"'"'{"temperature":0.7}'"'"''
```

PR #556, opened 2026-05-22
fix(ui): renderer local-escape + attribute-context CI lint

### LINKED ISSUE #553: Renderer URL attribute-boundary hardening (data-src reflection)
## Summary

Harden `shared_static/renderer.js`'s image-placeholder pattern, which reflects URLs into a `data-src` attribute without explicit escaping. An attacker-controlled image URL containing embedded double-quote characters could escape the attribute boundary and inject arbitrary HTML attributes (e.g., `onerror=`).

Pre-existing bug flagged out-of-scope during the interactive DOM-cleanup PR series (#532/#533/#534/#535). Surfacing here so it doesn't get memory-holed.

## The bug

In `turnstone/shared_static/renderer.js`, image-placeholder rendering constructs HTML like:

```javascript
data-src="' + url + '"
```

without HTML-attribute-escaping `url`. A URL such as:

```
https://example.com/x.png" onerror="alert(1)" data-x="
```

reflected into the template becomes:

```html
<img data-src="https://example.com/x.png" onerror="alert(1)" data-x="">
```

— now an `onerror` handler is injected.

## Threat model

- **Source of attacker control**: any path that allows untrusted URLs to reach renderer.js. Plausible sources: model output (markdown image references where the model writes the URL), MCP tool output that includes image URLs, user-pasted markdown containing image refs. Worth auditing all sources.
- **Impact**: XSS in the renderer's DOM context. Renderer runs in the main app origin, so the attacker gets full session cookie / API token access.
- **Mitigations today**: none specifically for `data-src`. CSP exists but does not cover inline `onerror` handlers added via attribute injection.

## Proposed scope

1. **Audit `shared_static/renderer.js`** for all attribute-construction sites, not just `data-src`. Any `attr="' + variable + '"` pattern is a candidate.
2. **Switch attribute-construction to DOM API** — use `element.setAttribute(name, value)` or `element.dataset.src = value`, which apply attribute-value escaping automatically. This matches the pattern landed in the interactive DOM-cleanup PR series for `innerHTML` migration.
3. **Add a lint rule** mirroring the cleanup PRs' `insertAdjacentHTML` lint: regex-match attribute-construction string-concat patterns in renderer files, fail CI on new instances. (The existing `\.innerHTML\s*=` lint pattern is the template.)
4. **Tests**: pin tests that pass attacker URLs through the renderer and assert no extra attributes/handlers materialize on the resulting DOM.

## Why 1.6.0

- Same release as the rest of the interactive DOM-cleanup follow-through (PR series #532-#535 landed under the 1.6.0aN umbrella).
- Security-class issues should be addressed promptly; deferring past 1.6.0 stable adds known-vuln baggage to the release.

## Risks / open questions

1. **Cross-renderer audit scope** — `renderer.js` is shared across interactive and coord UIs. The audit may surface other attribute-construction sites that are equally exposed but not currently flagged.
2. **CSP interaction** — a strict CSP (no inline scripts, no inline event handlers) would block the `onerror` injection regardless. Worth checking the current CSP posture and tightening it as defence-in-depth.
3. **Exploitability today** — currently I have not confirmed an end-to-end exploit path; the bug is structural. A spike to confirm or rule out an actual exploit chain is the first implementation step.

## Estimated lift

Half a day — single-file refactor + lint rule + a handful of pin tests. The cleanup PR series established the pattern; this follows the same template.

eous:
## Summary

- `inlineMarkdown`'s image and link renderers now `escapeHtml` each interpolated value (url, alt, label, domain) at the call site instead of relying on the upstream `escapeHtml(text)` pass at `renderer.js:13`. The current safety invariant is non-local — a future refactor calling those renderers from outside `inlineMarkdown` would silently regress. Locally-escaping makes the safety structural.
- New CI lint in `tests/test_renderer_js.py` scans `renderer.js` for `attr="' + ident` patterns; `ident` must be `escapeHtml(...)`, `safe*`-prefixed, or in the reviewer-approved `_RENDERER_KNOWN_SAFE_IDENTIFIERS` allowlist (currently just `"info"` for the `CALLOUT_TYPES` enum).
- Four pin tests use Python's `html.parser.HTMLParser` to verify attacker URLs and labels don't materialize event-handler attributes on rendered DOM.

**Defense-in-depth, not exploit fix.** The bug as originally described in #553 is not actively exploitable today because `inlineMarkdown` pre-escapes the captured URL/alt/label before the image/link regex runs. See the issue's spike comment for the boundary-spike finding. The local-escape and lint together prevent future regression if anyone calls those renderers from outside `inlineMarkdown`.

One known cosmetic effect: attribute values now double-escape (`&amp;quot;` instead of `&quot;`) since `inlineMarkdown` already escaped once. Functionally identical — the click handler at `renderer.js:93` calls `new URL(raw)` which rejects malformed URLs either way.

## Test plan

- [x] `pytest tests/test_renderer_js.py` — 67 passed (4.10s) including 4 new attacker-URL pin tests
- [x] `pytest tests/test_app_js.py` — 51 passed (1.91s, confirms existing `innerHTML` sink lint still green; renderer.js is intentionally outside `_UNSAFE_CODE_SINK_LINT_TARGETS` since it IS the trusted markdown→HTML transformer)
- [x] `ruff check tests/test_renderer_js.py` — clean
- [x] `/review` 4-finder pipeline — bug ✓, security ✓, perf ✓, quality found 2 (triplicate rationale + `#553` source-ref); verify confirmed 1, refuted 1 (PR/issue refs are established codebase pattern per spot-check); confirmed finding applied
- [ ] Optional manual smoke: render a markdown image with quote-in-URL in the browser, confirm placeholder displays and click-to-load rejects safely. Not strictly required since the Node-driven harness exercises the rendering path end-to-end.

Fixes #553.

### DIFF
--- a/tests/test_renderer_js.py
+++ b/tests/test_renderer_js.py
@@ -16,6 +16,7 @@ placeholder and not the raw delimiter.
 from __future__ import annotations
 
 import json
+import re
 import shutil
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


eous:
Thanks for the review — all 4 findings were valid and applied in 120d0320:

**C1 / C2 (double-escape on `&`)** — confirmed and reverted. Traced through: `alt` → inlineMarkdown escapes to `&amp;` → local `escapeHtml(url)` made it `&amp;amp;` → after attribute-parse + `getAttribute` returns literal `&amp;` → `new URL(...)` parses query as `a=1` and `amp;b=2` instead of `a=1` and `b=2`. Real functional regression, not just cosmetic. Switched the renderer to a convention-rename posture: regex callback params are now named `safeAlt` / `safeUrl` / `safeLabel` to signal the upstream-escape invariant, with no local re-escape. The attribute-context lint enforces the convention for future code, so defense-in-depth is preserved without the regression. Two added pin tests (`test_image_url_with_ampersand_not_double_escaped`, `test_link_url_with_ampersand_not_double_escaped`) pin the single-escape behavior.

**C3 (docstring/code mismatch on `safe[A-Z_]`)** — confirmed. The code only allowed `isupper()`. Dropped `_` from the docstring; JS uses camelCase anyway.

**C4 (`_all_attr_names` false-negative on bare `<script>`)** — confirmed. Refactored to `_parse_renderer_html` which returns `(start_tags, attr_pairs)` so attribute-less tag injections are visible. Updated `test_attacker_link_label_with_quote_renders_as_text` to check against `start_tags`.

All 69 renderer tests + 51 app-js tests green; ruff clean.

PR #687, opened 2026-06-22
fix(ci): exclude vllm-litellm deploy example from wheel-completeness check

eous:
#686 added `turnstone/deploy/vllm-litellm/` under the package dir, which trips the "all data files in wheel" CI check (those files aren't in the wheel) and is currently failing `main`.

They're a deploy **example** (compose + README + .env.example + Dockerfile): you clone the repo to run it; the package never references them. So they belong in the repo, not the wheel — added to the check's intentional-exclusion ALLOW list rather than shipping example docs in the wheel.

Verified locally: `python -m build --wheel` + the exact CI check → no missing files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,9 +113,16 @@ jobs:
             | grep -v '\.py$' | grep -v '\.dist-info' | grep -v '\.pyc' | grep -v '^File$' \
             | sort)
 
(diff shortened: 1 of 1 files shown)


eous:
Superseded by #688 — that PR carries this same wheel-check ALLOW fix alongside the 3-model example update (single PR avoids a ci.yml conflict between the two).

Issue #682, opened 2026-06-19
Docs: Entra pre-authorized clients for consent-free MCP OAuth (Azure)


eous:
## Summary

Write an operator runbook for **Azure / Entra** deployments that removes the per-server MCP OAuth consent prompt **today** — using Entra **admin consent** + **pre-authorized client applications** on Turnstone's existing `oauth_user` flow. No Turnstone code, no EMA / ID-JAG required.

This is the "available now" path for the Azure cohort, distinct from the standardized EMA / ID-JAG client work in #680 (which Entra does not support yet — see that issue's Entra reality check). It complements, rather than replaces, the per-user OAuth that ships today.

## Why

For trust-domain-aligned Entra deployments (Entra-protected internal MCP servers), the second consent dance after Turnstone login is gratuitous: the user already authenticated to Entra at Turnstone login. Entra can suppress the prompt via tenant configuration:

- **Tenant admin consent** for the downstream MCP server's delegated scopes, granted once for all users.
- **Pre-authorized applications** — list Turnstone's `client_id` on the MCP server's app registration (`preAuthorizedApplications`) for the required scopes, so Entra issues tokens without an interactive consent screen.

Result: a user signs into Turnstone via Entra OIDC (already supported), and the `oauth_user` MCP connection completes without a per-server consent prompt — purely tenant-side config.

## What to document (runbook)

1. Register / identify Turnstone's Entra app registration (the OIDC client).
2. On each MCP server's app registration: expose the API + delegated scopes; add Turnstone's `client_id` to **pre-authorized applications** for those scopes.
3. Grant **tenant admin consent** for the scopes.
4. Map the scopes onto Turnstone's per-server `oauth_user` config (`oauth_audience` / scopes).
5. Verify: fresh user → Turnstone Entra login → MCP server connects with no consent prompt.
6. Caveats: requires Global Admin / Privileged Role Admin; per-MCP-server app registration; works only within one tenant's trust domain; not a substitute for the cross-vendor EMA path (#680) for non-Entra IdPs.

## Scope

Documentation only — no code. Lands wherever the operator/admin docs live. Independent of the MCP v2 cluster (do anytime).

PR #706, opened 2026-06-23
Address remaining blockers to entra id on behalf of flow for user impersonation to mcp servers

metaclassing:
This is a collection of little snippets to resolve all the OBO flow problems required to get this talking to entra id for on behalf of user impersonating to protected mcp servers. we make sure turnstone checks these mcp servers on startup, and address some of microsofts opinionated implementations of oauth2/oidc and metadata provided by the identity provider.


eous:
This closes the functional gaps to enable entra id as described in https://github.com/turnstonelabs/turnstone/issues/682

PR #735, opened 2026-06-29
fix(memory): atomic single-statement upsert for memory save/update

eous:
## Problem

`save_structured_memory` upserted by *try INSERT → catch `IntegrityError` → SELECT + UPDATE*. On PostgreSQL, a model that saves the same key twice in a turn makes the second INSERT hit `uq_smem_name_scope`:

```
ERROR:  duplicate key value violates unique constraint "uq_smem_name_scope"
DETAIL:  Key (name, scope, scope_id)=(morning_tech_digest, global, ) already exists.
```

Postgres logs that at ERROR even when the fallback catches it, and the pattern threw + caught an exception on every update.

## Change

A new `StorageBackend.upsert_structured_memory` on both backends emits one atomic statement:

```sql
INSERT INTO structured_memories (...) VALUES (...)
ON CONFLICT (name, scope, scope_id)
DO UPDATE SET content = ..., description = ..., type = ..., updated = ..., last_accessed = ...
RETURNING <all columns>
```

- Returns `(row, was_update)` — the full saved row and whether an existing row was updated — like Django's `update_or_create`. `was_update` is the supplied (fresh) `memory_id` differing from the returned id. `save_structured_memory` is a thin wrapper over it.
- **Field semantics:** `description`/`mem_type` of `None` mean "leave unset" — the column default applies on insert and the stored value is kept on conflict; an explicit value (incl. `""` / `"general"`) overwrites. So clearing a description or setting type back to `"general"` now persists, where the prior `if mem_type != "general"` / `if description` semantics silently dropped it. An unrecognized type (a typo) is treated as unset (preserve) rather than coerced-and-written.
- The memory tool and the memories HTTP endpoint pass `None` for omitted fields and read effective `type`/`scope` from the returned row; the HTTP endpoint returns that row directly (one query — no follow-up SELECT).
- Removes the now-unused `update_structured_memory` primitive and its dead `STRUCTURED_MEMORY_MUTABLE` constant.

No `IntegrityError` round-trip, no constraint-violation log noise, race-safe under concurrent saves of the same key; `created` / `access_count` are preserved on conflict.

## Tests

- `tests/test_structured_memory_storage.py`: upsert-in-place (no raise, returns existing id), None-preserve vs explicit-overwrite, run on whichever backend `--storage-backend` selects — the save-over-existing path was previously SQLite-only (`tmp_db`).
- `tests/test_structured_memory.py`: facade `(row, was_update)` contract.
- `tests/test_session.py`: tool-path test (`_prepare_memory` → `_exec_memory`) — omit preserves, explicit overwrites, typo'd type preserved.
- mypy + ruff clean; full memory/storage/session/api suite green.

## Review

Reviewed via the multi-stage pipeline (bug / security / performance / quality finders + an earlier xhigh workflow pass). Security / performance / quality came back clean; one minor correctness finding — a typo'd `type` silently downgrading a stored type — was fixed and regression-tested.

### DIFF
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -3934,6 +3934,72 @@ class TestMemoryAccessTouch:
         )
         assert '<memory name="kafka_scaling"' in recomposed
 
+    def test_save_through_tool_preserves_omitted_overwrites_explicit(self, tmp_db):
+        """The None-sentinel flows through _prepare_memory -> _exec_memory: a
+        content-only re-save keeps the stored type/description, while an
+        explicit field overwrites it.  Guards the _prepare_memory omit->None
+        logic that the storage-level tests don't exercise."""
+        from turnstone.core.memory import get_structured_memory_by_name
+
+        session = self._empty_session()
+        item = session._prepare_memory(
+            "c1",
+            {
+                "action": "save",
+                "name": "digest",
+                "content": "v1",
+                "type": "reference",
+                "description": "daily digest",
+                "scope": "global",
+            },
+        )
+        assert "error" not in item
+        session._exec_memory(item)
+
+        # Content-only re-save (omits type/description) -> both preserved.
+        item2 = session._prepare_memory(
+            "c2", {"action": "save", "name": "digest", "content": "v2", "scope": "global"}
+        )
+        session._exec_memory(item2)
+        mem = get_structured_memory_by_name("digest", "global", "")
+        assert mem is not None
+        assert mem["content"] == "v2"
+        assert mem["type"] == "reference"
+        assert mem["description"] == "daily digest"
+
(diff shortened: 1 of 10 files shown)


eous:
@claude review this PR — the atomic single-statement upsert for structured-memory save/update. Please check: the upsert correctness (`INSERT ... ON CONFLICT (name, scope, scope_id) DO UPDATE ... RETURNING`), the None-vs-explicit field semantics (omit preserves the stored value, explicit `""`/`"general"` overwrites), the `(row, was_update)` derivation (resolved id != supplied fresh uuid), and the two callers (`_exec_memory` and the memories HTTP endpoint). Post your findings as a review.

PR #741, opened 2026-06-30
mcp dead transport fix and token refresh

metaclassing:
# MCP dead-transport: sessions never re-establish without a full worker restart

Branch: `fix/mcp-dead-transport-reconnect`
`turnstone/core/mcp_client.py` + regression tests in
`tests/test_mcp_client.py` and `tests/test_mcp_pool_auth_introspection.py`.
**Status:** THREE related fixes in one staged set (A dead-transport reconnect — covering idle
stream death, transport-gone, AND server-restart/stale-session; B OBO prime-refresh; C pool-aware
status) — implemented, unit-tested (all MCP suites green; 1 unrelated nginx:80 env failure),
deployed live to workers
---

## Symptom

A worker holds a long-lived `ClientSession` per MCP server. After a server's
streamable-http stream had been open for hours, tool calls to it began failing,
the per-server circuit breaker opened with exponential cooldown, and the agent
spun in a hung retry loop. **Only a full `turnstone-server` restart recovered it** —
refresh, reconnect, and re-sending a task all failed to re-establish the session.

Observed live on `storage3-monitoring` (a `static`/Bearer server so **not** an OAuth/OBO problem):

```
MCP circuit open for 'storage3-monitoring': 3 consecutive failures, cooldown 32s
MCP circuit open for 'storage3-monitoring': 4 consecutive failures, cooldown 64s
MCP circuit open for 'storage3-monitoring': 5 consecutive failures, cooldown 128s
Refresh failed for MCP server 'storage3-monitoring'
  File ".../core/mcp_client.py", line 2154, in _refresh_server_tools
    result = await session.list_tools()
  File ".../mcp/shared/session.py", line 281, in send_request
    await self._write_stream.send(...)
  File ".../anyio/streams/memory.py", line 218, in send_nowait
    raise ClosedResourceError
anyio.ClosedResourceError
```

The upstream endpoint was **100% healthy** the whole time — verified by a by-hand
MCP handshake + `health_overview` call straight to `http://storage3.secureobscure.com/mcp`,
which returned full ZFS/HBA health. The fault was entirely worker-side session state.

## Root cause

When the streamable-http transport's GET/SSE or POST stream dies (idle close,
keep-alive expiry, peer reset that the SDK's 1s GET-reconnect doesn't fully
recover), the **`ClientSession` object survives but its anyio write stream is
closed**. The session is a corpse. The two faces of that corpse are:

1. **`anyio.ClosedResourceError`** — raised directly when `list_tools()` /
   `call_tool()` writes to the closed in-memory stream (the traceback above).
2. **`McpError(CONNECTION_CLOSED)`** (`-32000`) — when the SDK's `post_writer`
   swallows the underlying transport error and surfaces a generic protocol error
   instead (see the module comment near the top of `mcp_client.py`).

**Every eviction/reconnect guard in the manager keyed off the wrong exception
family**, so neither corpse was ever recognized as "rebuild me":

| Site | Guard before the fix | What actually fires |
|---|---|---|
| `call_tool_sync` (dispatch) | evict session only on `BrokenPipeError \| ConnectionResetError \| EOFError` | `anyio.ClosedResourceError` |
| `_refresh_all` (periodic/manual refresh) | reconnect only when `state.session is None` | session is non-`None` but dead |
| `_classify_failure` (pool/OBO dispatch) | `isinstance(exc, McpError)` → `"protocol"` (treated as a *healthy* rejection) | dead transport surfaces as `McpError(CONNECTION_CLOSED)` |

Consequences:
- The dead session is **never set to `None`**, so `_cb_auto_reconnect` (which only
  fires on `session is None`) never runs.
- The periodic/manual refresh re-probes the **same corpse** every time and just
  logs "Refresh failed" — forever.
- `McpError(CONNECTION_CLOSED)` on the pool path was classified `"protocol"`, so it
  didn't trip the breaker and left the dead pool entry in place.

The code anticipated raw-socket errors (`BrokenPipeError`, etc.), which are what
the **stdio** transport raises — but the **streamable-http** transport raises the
anyio memory-stream errors, which slipped through every guard. Hence: full process
restart (fresh `_connect_all`) was the only recovery.

## The fix

A single shared classifier plus wiring it into the three session-use sites.

```python
import anyio  # httpx + mcp.types already imported; CONNECTION_CLOSED == -32000

def _is_dead_transport(exc: BaseException) -> bool:
    if isinstance(exc, anyio.ClosedResourceError | anyio.BrokenResourceError
                       | BrokenPipeError | ConnectionResetError | EOFError
                       | httpx.ConnectError | httpx.ConnectTimeout
                       | httpx.RemoteProtocolError):
        return True
    if isinstance(exc, McpError):
        if getattr(exc.error, "code", None) == mcp_types.CONNECTION_CLOSED:
            return True
        msg = (getattr(exc.error, "message", "") or "").lower()
        if "session terminated" in msg or "session not found" in msg:
            return True
    return False
```

Three death modes are covered (all "reconnect me", none a healthy protocol rejection):
1. **Local stream torn down** → `anyio.ClosedResourceError` / `BrokenResourceError` (idle close).
2. **Transport gone** → `McpError(CONNECTION_CLOSED)` (SDK-swallowed) or `httpx.ConnectError` /
   `RemoteProtocolError` (server down / peer closed mid-response).
3. **Server-side session lost** → the MCP server RESTARTED and dropped its session map, so our
   held `mcp-session-id` 404s; the SDK surfaces `McpError(code=32600, "Session terminated")`
   (`streamable_http._send_session_terminated_error`). Matched by message — 32600 isn't a
   documented MCP constant. Verified live: a stale id to storage3 returns HTTP 404
   `{"code":-32600,"message":"Session not found"}`; the SDK relays it as code 32600
   "Session terminated". WITHOUT this branch the dead session is reused → 404 forever → the
   restart-hang only a process restart clears. (This mode was missed by the first cut of the
   fix and added after a server-restart test exposed it.)

1. **`call_tool_sync`** — on `_is_dead_transport(exc)`: trip the breaker **and**
   evict `state.session = None` (so the next dispatch's `session is None` check
   fires `_cb_auto_reconnect` instead of re-using the corpse). Genuine `McpError`
   protocol rejections still don't trip the breaker.
2. **`_refresh_all`** — when `_refresh_server` raises a dead-transport error, null
   `state.session` so the reconnect branch at the top of the loop fires on the
   **next tick**. This turns the existing refresh into a **self-healing liveness
   probe** — which is the durability answer to "don't trust a stream that's been
   open for hours": prove-or-rebuild it on each refresh instead of assuming it's live.
3. **`_classify_failure`** — check `_is_dead_transport` **before** the
   `McpError → "protocol"` branch so `CONNECTION_CLOSED` is classified `"transport"`
   on the OBO/pool path too (trips the breaker, rebuilds the pool entry). `TimeoutError`
   remains `"transport"` as before.

Diff: `+60 / −4` in `turnstone/core/mcp_client.py`.

## Tests

Added to `tests/test_mcp_client.py::TestCircuitBreaker` (modeled on the existing
`test_connection_error_evicts_session` / `test_mcp_error_does_not_trip_circuit`):

- `test_closed_resource_error_evicts_session_and_trips_circuit`
- `test_connection_closed_mcperror_evicts_and_trips_circuit`
- `test_refresh_all_evicts_dead_session_so_next_tick_reconnects`

**These three FAIL on the clean branch** (reproducing the corpse-reuse: session
stays non-`None`, breaker doesn't trip / refresh never evicts) and **PASS with the
fix**. Full suite: **219 passed**; the one failure (`test_tcp_probe_default_port_http`)
is pre-existing and environmental — it expects `127.0.0.1:80` to refuse, but flowise
runs nginx on :80; it fails identically on the clean branch and is unrelated to this change.

The intent is safety: a transient refresh failure during a bulk prime must never revoke a
token and force re-consent. But it creates a **chicken-and-egg for any expired token**:

1. access token expired → prime **skips** the server →
2. its tools never populate the per-user catalog (`get_tools(user_id)`) →
3. the model can't see the tool, so it never emits a call →
4. **lazy dispatch — the ONLY path that refreshes (refresh-on-401) — never fires** →
5. the (still-valid) refresh token is never exercised → permanently stuck "connecting" / "no token".

Confirmed against `mcp_user_tokens.expires_at` (user `549bb115…`):

## Fix B — prime drives a guarded refresh (IMPLEMENTED)

In `_prime_user_pools._prime_one`, the non-refreshing read + near-expiry skip is replaced by a
call to the SAME guarded resolver the lazy-dispatch path uses,
`get_user_access_token_classified(...)`. It refreshes an expired/near-expiry token and persists
it, but a TRANSIENT AS/network failure keeps the token (`kind=refresh_failed_transient`, no
revoke) — only `kind == "token"` warms the pool. This closes the chicken-and-egg: an expired
token is now refreshed at session start instead of stranding the pool forever. The
no-revoke-on-transient-failure safety property the original skip protected is preserved by the
resolver's failure classifier (only a permanent / sustained-ambiguous rejection clears a token,
which requires re-consent anyway).

## Fix C — pool-aware status (IMPLEMENTED)

`get_server_status` now special-cases `oauth_user` servers (`_oauth_user_server_status`):
reports `connected=True` with a representative catalog count and `user_pools=N` when ≥1 user has
a warm pool entry, instead of the static-session check that always read "connecting".
`get_all_server_status` unions in `_oauth_user_server_names` (absent from `_server_configs`) so
the console list receives real per-user status for them. (Confirm the console pill flips after a
pool warms — the worker now reports it; a stale UI cache, if any, is the only remaining variable.)

## Tests (B + C)

In `tests/test_mcp_pool_auth_introspection.py`: the obsolete
`test_prime_user_pools_skips_near_expiry_without_revoking` is replaced by
`test_prime_user_pools_refreshes_expired_token_and_warms` (expired token → refreshed → warmed)
and `..._transient_refresh_failure_skips_without_revoking` (transient → not warmed, token kept);
plus `TestOAuthUserServerStatus` (connected-when-warm / connecting-when-cold). All existing pool
tests still pass.

## End-to-end validation (live)

Triggered ChatSession for the affected user via the route API. Worker logged
`mcp pool auto-primed at session start` for **all four** oauth_user servers including the two
previously-stuck ones; `mcp_user_tokens` then showed Azure-CLI + Cloudflare with
`last_refreshed = 03:03:01` and `expires_at` advanced ~1h into the future (refresh succeeded +
persisted). The soak check reports zero flags; no `ClosedResourceError` / circuit-open / stuck
servers on the worker.

### DIFF
--- a/tests/test_mcp_client.py
+++ b/tests/test_mcp_client.py
@@ -2487,6 +2487,133 @@ class TestCircuitBreaker:
         # Circuit should NOT have recorded a failure
         assert mgr._consecutive_failures.get("test", 0) == 0
 
+    def test_closed_resource_error_evicts_session_and_trips_circuit(self):
+        """Regression: the MCP SDK's streamable-http transport raises
+        ``anyio.ClosedResourceError`` (NOT BrokenPipeError) when its write
+        stream is dead. That must evict the session AND trip the breaker —
+        otherwise the corpse session is re-used on every call forever and
+        only a full process restart recovers it."""
+        import anyio
+
+        mgr = MCPClientManager({"test": {"type": "stdio", "command": "echo"}})
+        mock_session = MagicMock()
+        mock_session.call_tool = MagicMock(return_value="sentinel")
+        _seed_static_state(mgr, "test", session=mock_session)
+        mgr._loop = MagicMock()
+        mgr._tool_map["mcp__test__ping"] = ("test", "ping")
+        mock_future = MagicMock()
+        mock_future.result.side_effect = anyio.ClosedResourceError()
+        with (
+            patch("asyncio.run_coroutine_threadsafe", new=_dispatch_stub(mock_future)),
+            pytest.raises(anyio.ClosedResourceError),
+        ):
+            mgr.call_tool_sync("mcp__test__ping", {}, timeout=5)
+        assert mgr._static_servers["test"].session is None
+        assert mgr._consecutive_failures.get("test", 0) == 1
+
+    def test_session_terminated_mcperror_evicts_and_trips_circuit(self):
+        """Regression: when the MCP SERVER restarts and loses its session map, our
+        held mcp-session-id is stale; the server returns HTTP 404 and the SDK
+        surfaces McpError(code=32600, 'Session terminated'). That is NOT a healthy
+        protocol rejection — the session must be evicted so the next dispatch
+        reconnects with a fresh initialize; reusing it 404s forever (restart-hang)."""
+        from mcp import McpError
+        from mcp.types import ErrorData
+
+        mgr = MCPClientManager({"test": {"type": "stdio", "command": "echo"}})
+        mock_session = MagicMock()
+        mock_session.call_tool = MagicMock(return_value="sentinel")
+        _seed_static_state(mgr, "test", session=mock_session)
+        mgr._loop = MagicMock()
+        mgr._tool_map["mcp__test__ping"] = ("test", "ping")
+        mock_future = MagicMock()
+        # Exactly what the streamable-http SDK injects on a 404 stale session.
+        mock_future.result.side_effect = McpError(
+            ErrorData(code=32600, message="Session terminated")
+        )
+        with (
+            patch("asyncio.run_coroutine_threadsafe", new=_dispatch_stub(mock_future)),
+            pytest.raises(McpError),
+        ):
+            mgr.call_tool_sync("mcp__test__ping", {}, timeout=5)
+        assert mgr._static_servers["test"].session is None
+        assert mgr._consecutive_failures.get("test", 0) == 1
+
+    def test_httpx_connect_error_evicts_session(self):
+        """A dead underlying httpx connection (server down mid-call) is transport
+        death, not a protocol rejection — evict so the next call reconnects."""
+        import httpx
+
+        mgr = MCPClientManager({"test": {"type": "stdio", "command": "echo"}})
+        mock_session = MagicMock()
+        mock_session.call_tool = MagicMock(return_value="sentinel")
+        _seed_static_state(mgr, "test", session=mock_session)
+        mgr._loop = MagicMock()
+        mgr._tool_map["mcp__test__ping"] = ("test", "ping")
+        mock_future = MagicMock()
+        mock_future.result.side_effect = httpx.ConnectError("connection refused")
+        with (
+            patch("asyncio.run_coroutine_threadsafe", new=_dispatch_stub(mock_future)),
+            pytest.raises(httpx.ConnectError),
+        ):
+            mgr.call_tool_sync("mcp__test__ping", {}, timeout=5)
+        assert mgr._static_servers["test"].session is None
+        assert mgr._consecutive_failures.get("test", 0) == 1
+
+    def test_connection_closed_mcperror_evicts_and_trips_circuit(self):
+        """Regression: when the SDK's ``post_writer`` swallows the transport
+        error, a dead connection surfaces as ``McpError(CONNECTION_CLOSED)``.
+        Unlike a genuine protocol rejection, this MUST evict + trip the
+        breaker so the next dispatch reconnects instead of looping."""
+        from mcp import McpError
+        from mcp.types import CONNECTION_CLOSED, ErrorData
+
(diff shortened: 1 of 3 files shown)


eous:
@claude Please /code-review max this pull request

PR #767, opened 2026-07-04
feat(mcp): proactively keep consented OAuth (OBO) tokens fresh for unattended work

eous:
## Problem

Per-user OAuth (`auth_type=oauth_user`, on-behalf-of) token refresh is entirely **lazy**: a token is refreshed only when a tool is dispatched or a session binds the acting user, and a dead refresh token is discovered only when a dispatch *fails*. That assumes a human is driving the session — which breaks precisely for **autonomous / scheduled work** acting on behalf of an absent user:

- the access token may be expired (first dispatch pays the refresh latency),
- it may be in a transient-failure cooldown (tool unavailable), or
- the grant may be dead, with nobody present to re-consent.

The only periodic MCP-loop task today (idle eviction) actively **tears OBO connections down**; nothing keeps tokens warm or surfaces dead grants ahead of the work that needs them.

## Change

A background token-freshness sweep (`_user_token_sweep_loop`, default 240s) that keeps every consented `oauth_user` grant hot **without keeping connections warm** and **without mutating consent state on a timer**.

- **Strictly `oauth_user`-scoped.** Gates on `_oauth_user_server_names` and drives off `mcp_user_tokens` rows. A static / no-auth server has neither, so it is *structurally* invisible to the sweep — zero DB scans, zero authorization-server round-trips, zero MCP-server calls. Never connects to the MCP server; connections stay lazy (one cheap reconnect on first dispatch).
- **Observe-only.** A new `revoke_on_failure=False` parameter on `get_user_access_token_classified` makes the background path non-destructive: a timer never deletes a token, emits `token_revoked`, or mutates the shared ambiguous-streak / cooldown. A dead grant is only *surfaced* (proactive dashboard pending-consent badge); the authoritative revoke stays on the lazy-dispatch path where a real user action justifies it. Because the row survives, a spurious server-wide `invalid_grant` (an AS maintenance window) **self-heals** — the badge is dropped on the tick the grant works again.
- **Keepalive.** Force-refreshes a grant whose refresh token has sat un-exercised past `user_token_refresh_keepalive_seconds` (default 1800s) even while the access token is still fresh, so a provider that ages out *idle* refresh tokens can't expire one between a user's real sessions.
- **Robust surfacing.** Dead grants are surfaced once per transition, pinning the pair only after a durable badge write so a failed persist retries rather than being lost. The first sweep runs after a short startup grace, so a restart surfaces a grant that died during downtime within seconds.

Cadence `<= 0` disables the sweep; a positive value is floored (30s) so a misconfigured tiny cadence can't busy-loop. The sweep reuses the per-key refresh lock, so a keepalive force can't double-refresh against a concurrent dispatch.

### Storage
Adds `list_mcp_user_token_reconcile_targets()` → `(user_id, server_name, COALESCE(last_refreshed, created))` (expiry-unfiltered, no ciphertext projected) on the protocol, sqlite, and postgres backends.

## Deliberately out of scope
- **Connection pre-warming** (keeping OBO sockets hot for zero first-dispatch latency) — connections re-establish lazily and cheaply; the token is what needs keeping fresh.
- **Static-server auto-reconnect** (self-healing a downed non-OBO server without manual intervention) — a real, separate gap being spiked next.
- **Per-node sharding** of the sweep — a scaling concern only at large fleet sizes; a follow-up if OBO ever scales past low-tens of consented users per node.

## Tests
No-auth invisibility, observe-only non-destruction (token kept + shared streak untouched on a background permanent / ambiguous failure), keepalive gating (due / not-due / disabled / unparseable), badge persist-then-pin retry, self-heal on recovery, cadence clamp / disable, and the storage enumerator. Full MCP / OAuth / storage / session suites pass locally.

### DIFF
--- a/tests/test_mcp_client.py
+++ b/tests/test_mcp_client.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import concurrent.futures
+import contextlib
 import inspect
 import json
 import time
@@ -2180,6 +2181,13 @@ class TestConnectOneUnreachable:
 
         loop = asyncio.new_event_loop()
         loop.run_until_complete(mgr._connect_all())
+        # _connect_all spawns the long-lived token-freshness sweep task; cancel
+        # it before closing the loop so it isn't destroyed while pending.
+        sweep = mgr._user_token_sweep_task
+        if sweep is not None:
+            sweep.cancel()
+            with contextlib.suppress(BaseException):
+                loop.run_until_complete(sweep)
         loop.close()
 
         bad_state = mgr._static_servers.get("bad")
--- a/tests/test_mcp_oauth_refresh.py
+++ b/tests/test_mcp_oauth_refresh.py
@@ -408,6 +408,97 @@ class TestRefreshFailureClassification:
         assert ("user-1", "srv-oauth") not in state.mcp_oauth_refresh_locks
 
 
+class TestObserveOnlyLookup:
+    """``revoke_on_failure=False`` (the background token-freshness sweep): still
+    refresh a healthy token, but on failure NEVER delete a token or mutate the
+    shared streak — a timer must not destroy consent or move a foreground user's
+    revoke threshold. A permanent rejection surfaces as ``refresh_failed`` with
+    the row INTACT; an ambiguous one as transient with the streak untouched."""
+
+    def _lookup(self, state: SimpleNamespace) -> Any:
+        from turnstone.core.mcp_oauth import get_user_access_token_classified
+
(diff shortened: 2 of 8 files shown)


eous:
@claude Investigate copilots finding "Docstring claims the dispatch path passes last_ws_id / last_tool_call_id through a wrapper, but _write_pending_consent() always writes both as None and _record_pending_consent_best_effort() does not plumb any ids. This is misleading for future maintenance (it suggests data is being captured when it is not)." Is this a docstring issue or a gap in our implementation?

Issue #865, opened 2026-07-17
Send-worker error-surface hygiene + completion-notify honesty (per-turn error-recorded signal)


eous:
## Summary

The initial-message worker fix settles a failed first turn to `state=error` with a readable `last_error` via a new `ChatSession.ensure_error_recorded`, but scopes that primitive to the **fresh-session** init worker only. This issue batches the deferred, related error-surface hygiene across the other send-worker closures and the completion-notification surface, to be redesigned once at a shared chokepoint.

## Why the primitive is fresh-session-only

`ensure_error_recorded`'s idempotency guard (`_has_persisted_error`) is **session-lifetime** — it is cleared only by an idle/running state emission, not per turn. So it is correct only for a first-turn/fresh session (the init worker). A **session-reuse** caller that routed through it — e.g. a `/retry` after a prior errored turn, failing in `send`'s pre-`try` section — would find the flag still stale-True, no-op, and **swallow the fresh error** while surfacing the stale prior `last_error`. That is why the retry closure is deliberately left on its pre-fix path rather than routed through the recorder.

## Structural fix (one chokepoint)

Replace the session-lifetime `_has_persisted_error` idempotency with a **per-send error-recorded signal** (reset at the top of `send`, set by `_record_fatal_error`). An idempotent recorder built on that is safe for reuse callers, which closes the swallow *and* every pre-try hole below at a single place.

## Pre-try holes to close

Each of these drops `state=error` / `last_error` when `send` raises before its `try` (model-registry refresh, user-turn append, system-message recompose). Because the coordinator's `wait`/`inspect` reads `last_error` only for `state=='error'`, the failure is invisible:

- **retry closure `_run`** (turnstone/server.py) — currently on its pre-fix path: raw-exception UI leak, double `state=error` emit, and pre-try no-persist (surfaces a stale `last_error`).
- **main `/send` worker** (turnstone/core/session_routes.py:4130) — `except Exception` does `on_stream_end` only, assuming `send` recorded.
- **coordinator send** (turnstone/console/coordinator_adapter.py:349) — logs only.
- **wake worker** `deliver_wake_nudge_from_queue` (turnstone/core/session.py:11507) — catches only `GenerationCancelled`; a pre-try `Exception` propagates to `session_worker._runner`, which only logs, so the wake silently no-ops with no error surfaced.

## Completion-notification honesty (external-channel surface)

For scheduled/unattended workstreams the notification is the only failure surface, and it is currently outcome-blind:

[...]

PR #866, opened 2026-07-18
fix(server): init-message worker exits to error, not idle, on first-turn failure

eous:
## Problem

The initial-message worker (`_run_initial`) collapsed cancel and backend-error exits into one `except (Exception, GenerationCancelled)` arm that always stamped `state=idle`, clobbering the `state=error` that `session.send`'s `_record_fatal_error` had just persisted. A spawned child whose first turn hit a backend failure (unreachable model server, exhausted quota, auth error) therefore read as an empty, successful turn — the coordinator's `wait_for_workstream` / `inspect_workstream` surfaces `last_error` only for `state=='error'` — so the real failure surfaced only after a manual nudge re-ran the turn and hit the same error synchronously.

Surfaced during live testing: a coordinator spawned children, they came back idle/"(no recent assistant output)", and it took repeated nudges to discover the backend was down.

## Fix

- **Split the arm** — cancel -> `idle`, exception -> `error`. The failed child settles at `state=error` and the first `wait_for_workstream` returns the enriched backend error inline. Same latent bug fixed for scheduled tasks (same endpoint/closure).
- **New idempotent `ChatSession.ensure_error_recorded`** — a no-op when `send` recorded the error in-line (the common backend path, no duplicate state emit) and the recorder when a pre-try exception (`_refresh_model_from_registry`, `_append_user_turn`, `_init_system_messages`) bypassed `send`'s own handler, so `state=error` always carries a meaningful `last_error`. Its `_has_persisted_error` guard is session-lifetime, so it is scoped to `_run_initial`'s fresh first-turn session only; the docstring spells out why a session-reuse caller (retry / main send / coord send / wake) must not route through it until the per-turn error-recorded signal in #865 lands.
- **A failed first turn is deliberately terminal for automated wakes** — a non-ready error terminal, not the idle ready-set that timer/watch wakes recur to; explicit user/coordinator action reactivates it (documented at the site).
- **Stable model identifier in the error** — the enriched backend error now leads with the model *alias* the coordinator references everywhere (`list_nodes`, spawn) and annotates the backend id for the operator: `model=DeepSeek-V4-Flash (id=deepseek-v4-flash)` (collapsing when they coincide). A model routing around a failed model now correlates it against those surfaces without a lookup, instead of spending reasoning tokens reconciling the alias against a backend id it never sees anywhere else.

## Tests

`TestInitialWorkerFailureState` asserts the coordinator-visible manager state + persisted `last_error`, recorded **exactly once**, across the common-backend and pre-try error legs, plus cancel-to-idle — each gated on worker completion (not a default state, which would pass vacuously). New `_format_backend_error` tests pin the alias/id label (both shown when they differ, collapsed when equal).

## Deferred to #865

The completion-notification honesty surface (a failed/cancelled init still notifies the "(Task completed)" fallback) and the error-recording hygiene of the other send-worker closures (retry / main send / coord send / wake) — the per-turn error-recorded signal that makes `ensure_error_recorded` reuse-safe.

### DIFF
--- a/tests/test_server_attachments_on_create.py
+++ b/tests/test_server_attachments_on_create.py
@@ -9,6 +9,7 @@ Exercises:
 
 from __future__ import annotations
 
+import contextlib
 import json
 import queue
 import threading
@@ -16,6 +17,8 @@ import time
 
 import pytest
 from starlette.testclient import TestClient
+
+from tests._helpers import wait_until
 
 # Magic-byte-valid 1x1 PNG (matches the fixture in test_server_attachments_endpoints.py)
 PNG_1x1 = (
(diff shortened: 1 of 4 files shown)


eous:
Copilot's credential-leak finding on the retry closure (server.py:736) is addressed in 55d346cd: the display string is now sanitized inline via `sanitize_error_text` — the same redactor `_record_fatal_error` uses, so a `user:pass@host` base-URL no longer reaches the dashboard SSE. The separable recording hygiene on that closure (double `state=error` emit, pre-try no-persist) stays tracked in #865.

PR #898, opened 2026-07-23
Per-user oauth2 on behalf flow for model calls

metaclassing:
As per usual, possibly something you want to hand your bots to implement differently. This is what I have deployed a couple places now and works well with model routers / llm gateway providers that do OAUTH instead of static api keys.

---

### Summary
Lets a model backend authenticate to an Entra-fronted LLM gateway with a
**per-request token minted from the deployment's own `[oidc]` app registration**
instead of one shared static API key — so the gateway can attribute calls to the
actual user (or to the app as a managed identity) rather than a single key.
Opt-in per model alias; `static` stays the default and existing rows are untouched.

### What's added — a per-alias `auth_mode`
- **`static`** (default) — send the stored `api_key`. Unchanged behaviour.
- **`entra_obo`** — mint a **per-user** On-Behalf-Of token for `obo_audience`
  from the caller's captured refresh credential (`oidc_user_credentials`); the
  gateway sees the human's identity.
- **`entra_app`** — mint an **app-identity** token via the client-credentials
  grant from the deployment's `[oidc]` client id + secret (no user; one shared
  machine identity). Also serves userless turns (utility / coordinator / service
  / CLI) that OBO can't.

### How it works
- Two additive columns on `model_definitions` (`auth_mode`, `obo_audience`) via
  **migration 068** (tested downgrade; existing rows default to `static`).
- `mint_obo_access_token` / `mint_app_access_token` (`mcp_oauth.py`) reuse the
  existing OBO grant legs, token-endpoint POST helper, refresh-token rotation
  CAS, and the `mcp_user_tokens` mint-cache — keyed under synthetic
  `__model_obo__:<audience>` / `__model_app__:<audience>` rows so a minted token
  is shared across worker nodes and inspectable.
- At the call site the token is bound as the SDK client's credential via
  **`client.with_options(api_key=…)`** — a copy that reuses the connection pool
  but swaps the credential, so the SDK emits it as its own auth header
  (`x-api-key` for Anthropic, `Authorization: Bearer` for OpenAI-style). `None`
  (static alias / no user / failed mint) falls back to the client's static key.

### Design note — why `with_options`, not header injection
An earlier revision injected the token via `create_streaming(extra_headers=…)`.
That works for the OpenAI SDK, but the **Anthropic SDK does not let
`extra_headers` override its `x-api-key`**, so anthropic backends silently
shipped the placeholder credential and the gateway returned 401. Binding via
`with_options(api_key=…)` uses each SDK's own auth path, which is why the
provider→header mapping helper could be deleted entirely (net −code).

### Config / compatibility
- Opt-in per alias in the admin model dialog (Auth mode + OBO audience fields).
- **No new required env** — reuses the existing `[oidc]` confidential-client
  config. `obo_audience` is the resource; the mint requests `<audience>/.default`.
- Fully backward compatible: `static` default, additive migration, no behaviour
  change unless an alias opts in.

### Testing
`tests/test_model_provider_obo.py` covers: migration up/down/roundtrip, storage +
admin-load round-trip, both mints (happy / DB-cache / rotation / force-refresh /
no-credential / oidc-disabled / rejected-grant), the call-site token resolve
(static / no-user / failed-mint fallbacks; provider-agnostic), and `entra_app`
(mints with zero user context). `ruff`, `mypy --strict`, and pytest all pass.

### Out of scope
The gateway-side JWT validation / identity→account mapping is external
configuration. `entra_app` is a separate explicit mode (not an implicit OBO
fallback) to keep identity semantics predictable.


eous:
Just kicked off the same local code-review pipeline my stuff goes through

eous:
Thanks for this — and for the write-up, which made the review much faster. Sending it through a full multi-pass review plus a dataflow trace of the proposed fixes before anyone starts editing.

**The design is the good kind of boring, and that's a compliment.** You reused the existing OBO grant legs, the refresh-token rotation CAS, the cluster advisory lock and the `mcp_user_tokens` mint-cache instead of forking parallel copies — that's exactly right, and it's why the diff is as small as it is for what it does. The `with_options(api_key=…)` binding is the correct mechanism and your design note explaining why `extra_headers` can't work for the Anthropic SDK is preserved in the docstring where the next reader will find it. Migration 068 is additive with a tested downgrade, and 654 lines of tests is a serious contribution.

The findings below are **23 items: 8 significant, 14 smaller, 1 optional.** No criticals. Every one was re-read against source by an independent pass, then the whole fix set was traced through the codebase — which turned up several places where the obvious fix doesn't work as stated. Those corrections are folded in below, because they're the difference between "one-line fix" and "not one line".

Happy to take any subset of this ourselves if you'd rather not carry it all — say the word.

---

## Before merge

### 1. The mint is only reachable on one of four hosts

`_model_obo_token` (`core/session.py:6006-6008`) resolves the token through `self._mcp_client`, and bails silently when that's `None`. It's `None` in more situations than you'd expect:

- `create_mcp_client` returns `None` outright when no MCP servers are configured (`core/mcp_client.py:9176-9177`). A deployment that wants per-user model auth but has no MCP servers gets nothing.
- `session.py:1809` — `self._mcp_client = mcp_client if self._persona_mcp else None`. A persona with the MCP lever off silently disables model auth, and `_drop_mcp_surface()` (`:3226`) does the same mid-session when a `resume()` adopts an MCP-off stamp, so a working session can flip to 401s partway through.

Worse, the sync bridges also bail on `self._app_state is None` (`mcp_client.py:1006`, `:1037`), and `set_app_state` is called in exactly one file:

```
turnstone/server.py:3880   mcp_mgr.set_app_state(request.app.state)
turnstone/server.py:5923   mcp_client.set_app_state(app.state)
```

`cli.py:1393` sets storage but not app_state; `console/server.py:5219` and `:10696-10699` set neither. **So `entra_obo`/`entra_app` is node-only today** — CLI turns and console-hosted coordinator workstreams can never mint, and none of them log why. In every one of these cases `get_client` has already built the SDK client with `api_key="obo-placeholder-unused"` (`model_registry.py:163-165`), so the symptom is an opaque 401 with nothing in the log.

**Fix:** resolve the credential from a reference that is neither persona-gated nor conditioned on MCP servers existing — an ungated `self._mcp_mint_client` captured at construction from the raw kwarg (the persona gate governs *tool surface*, not model auth). Land the warning first and unconditionally: a log line when `cfg.auth_mode` is an entra mode but the mint client is unavailable is a four-line edit that converts every one of these from a silent 401 into something diagnosable.

One thing we couldn't settle from the code: **is the console's `MCPClientManager` deliberately app-state-less?** Wiring it would newly enable per-user OBO *and* `oauth_user` MCP pools for coordinator sessions — a behaviour change well beyond this PR. Your call, and worth a comment either way.

### 2. Neither mint has a failure cooldown, so a dead grant retries every turn

`mint_obo_access_token`'s entire failure handling is `except MCPOAuthRefreshFailed: log.warning(...); return None` (`core/mcp_oauth.py:2829-2831`); the credential-missing (`:2787-2789`) and missing-`access_token` (`:2834-2836`) returns are the same shape. The classified sibling you modelled this on has both a `_refresh_in_cooldown` gate (`:2431`) and `_arm_cooldown` — its comment spells out the reason.

So an expired client secret or a missing admin consent means: every model turn takes the in-process lock plus a Postgres advisory lock (whose `_PgRefreshLock.__aenter__` spawns a fresh ThreadPoolExecutor and spins to a 30s timeout), POSTs the IdP, waits 10s, and fails. `_model_obo_token` runs per turn (`session.py:5498`, `:5160`, `:16147`), so a 20-tool-round turn is 20+ IdP round-trips. For `entra_app` the advisory key is the constant `__app__` pseudo-user, so it serializes **cluster-wide per audience**.

**Fix:** gate on `_refresh_in_cooldown` before taking either lock, `_arm_cooldown` on all three failure returns, `_clear_refresh_backoff` on all three success returns. Two placement details that matter:

- The gate belongs *after* the pre-lock cache serve (`:2764`) and *before* `_refresh_lock_for` (`:2771`), so a cooldown never denies a warm cached token — that's what the classified path's `if not fresh and _refresh_in_cooldown(...)` encodes.
- **Key it on `cache_server`, never on `credential_key`.** `credential_key = "__obo__:<issuer>"` (`:2772`) is the *same* key the classified MCP mint takes at `:2509-2511` and gates on at `:2431` — arming there would let one broken model alias suppress every `oauth_obo` MCP tool dispatch for that user. Worth stating as an invariant in the code.

### 3. Deprovisioning doesn't reach the model-gateway bearer

`admin_delete_oidc_identity` purges cache rows for `obo_server_names(storage)` only (`console/server.py:5768-5786`), which enumerates `mcp_servers` rows with `auth_type == "oauth_obo"`. MCP server names are validated `^[a-zA-Z0-9._-]+$`, so that set can *never* contain `__model_obo__:https://gateway`. Meanwhile `mint_obo_access_token` serves its cache row (`mcp_oauth.py:2755-2763`) before it ever reads the credential (`:2784`) — so after an unlink, the departed user's scheduled and channel-driven runs keep calling the model gateway with their live bearer until the token expires. The comment at `server.py:5747-5753` already flags this exact hazard for the MCP rows.

**Fix:** purge `__model_obo__:`-prefixed rows alongside the existing loop. Two corrections to the obvious version:

- Match `__model_obo__:` **only**, not a shared prefix tuple. `__model_app__:` rows live under the `__app__` pseudo-user and are a shared machine identity, not the departing user's.
- The handler's surrounding block works through `token_store`, so `token_store.list_user_token_metadata(user_id)` (`core/mcp_crypto.py:489`) is the symmetric call; the counter is `obo_cache_purged` (`:5756`).

Related and worth knowing: `_APP_CACHE_USER = "__app__"` is immune to both the `delete_user` cascade and this purge by design — so **the only revocation lever for a leaked `entra_app` token today is waiting out `expires_at`.** That's defensible, but it should be a documented property rather than an emergent one.

### 4. `obo_audience` is unconstrained, on an endpoint gated below the equivalent MCP capability

The only validation is presence (`console/server.py:11698-11706`). The value flows unchecked into `scope=f"{audience}/.default"` against the user's stored refresh token, and the minted token is bound as the SDK credential sent to the admin-set `base_url`. So an alias with `base_url=https://attacker.example`, `auth_mode=entra_obo`, `obo_audience=https://graph.microsoft.com` redeems every user's captured IdP credential for Graph and ships the result to a host of the caller's choosing.

The handler is gated on `admin.models` (`:11618`, `:11783`); writing `oauth_audience` on an MCP server — the same capability — sits behind `admin.mcp` (`:10079`). These have no hierarchy, so this crosses a permission boundary.

**Fix:** an operator-configured allow-list for the audience, plus `admin.mcp` on writes that set or change these two fields. Three corrections:

- **Allow-list, not a host-match against `base_url`.** Entra resource identifiers are commonly `api://<guid>` App ID URIs that share no host with the gateway — your own `ModelConfig` comment calls it "the resource App ID URI" — so a host-match would reject legitimate configs.
- **Gate on the effective value *changing*, not on the key being present in the body.** `console/static/admin.js:7712-7717` sends both fields unconditionally and the same submit function does POST and PUT, so "if the key is in the body" would make a temperature tweak require `admin.mcp`. Evaluate after the merge at `:11901-11902`, before the write at `:11909`.
- `allow_service_bypass=False` would be *stricter* than the MCP twin (which uses the default), not equal to it. Probably right for a capability-escalating field, but it's a deliberate hardening step — and we couldn't rule out an external automation client provisioning model definitions with a service token, which it would break.

### 5. Five model lanes never receive the token

`obo_api_key` reaches `model_turn` from exactly two places:

```
turnstone/core/session.py:5160    obo_api_key=self._model_obo_token(lane.alias),   # utility completion
turnstone/core/session.py:16147   obo_api_key=self._model_obo_token(lane.alias),   # sub-agent
```

It's omitted at `judge.py:1410`, `perception.py:125`, `output_guard_judge.py:555`, `eval/core.py:339` and `optimizer.py:301/451/771/909/1050`. Each resolves its client through the registry, so on a keyless alias they all send `obo-placeholder-unused` and 401. `mint_app_access_token`'s docstring advertises coverage of exactly the "utility / coordinator / service / CLI turns" these represent.

Scope this honestly: `entra_obo` genuinely *can't* work in these lanes (there's no user), and `_model_obo_token` already handles that correctly. **The mode that's actually broken here is `entra_app`.**

Two corrections to the fix:

- **"Resolve it inside `model_turn` from the lane" isn't implementable.** `ModelLane` (`model_turn.py:338-376`) carries `registry` and `alias` but no `app_state`, no manager, and no user id; `model_turn` is a plain sync function and there's no manager singleton. It can *read* `auth_mode` but it can't mint. The workable shape is a resolver callable on the lane, supplied at `resolve_lane` time — which moves the forgetting risk to the 12 `resolve_lane` sites rather than eliminating it.
- **The optimizer sites can't be threaded at all yet**: those lanes are built registry-less (`resolve_lane(prov, client, model)` at `optimizer.py:246/452/768/910/1051`, no alias, no registry). Passing a token there is a no-op until the optimizer gets a registry and an alias.

For the regression guard, prefer a behavioural assertion — *a lane configured `entra_app` never issues a call carrying `obo-placeholder-unused`* — over one that enumerates `model_turn(` call sites, since the behavioural one survives a new lane being added.

**Open question we can't answer from the code:** are the judge / output-guard / perception lanes *supposed* to run under the caller's delegated authority? `entra_app` is the obviously-safe answer; `entra_obo` is not obviously safe — the judge reads tool arguments and trajectory content. Note also that the perception memo is keyed `(alias, content_hash)` with no user component, so a per-user token there would leak one user's authorized output to another.

### 6. The two load-bearing behaviours have no test

- **The `with_options` binding** (`model_turn.py:749`, mirrored `session.py:5500`) — the one step your description names as the cause of a production 401 — has no test. `grep obo_api_key tests/` returns nothing, and `tests/test_model_turn.py` builds every lane with `client=object()`, which has no `with_options`, so the branch is provably unreachable from the suite. The MagicMock shape you need already exists at `tests/test_model_registry.py:1616`. Two gotchas: `_lane()` (`:74-76`) hardcodes `client=object()`, so passing `client=` through `**kw` raises `TypeError` — construct `ModelLane(...)` directly; and `with_options` is called *inside* the drain-retry loop at `:740`, so `assert_called_once_with` is only sound against a provider that doesn't trigger a drain retry.
- **The four 400-validation branches** (`console/server.py:11699-11706`, `:11892-11907`) have no test — including the post-merge cross-field check, whose whole purpose is a request touching one half of the pair. `tests/test_admin_model_registry_refresh.py` already has the `_make_client` harness and asserts 400s on bad `api_surface` at `:1005/1031/1066`. Four cases: `auth_mode="bogus"` → 400; `entra_obo` with no audience → 400; PUT `{"auth_mode": "entra_obo"}` onto a row with empty stored audience → 400; PUT `{"obo_audience": ""}` onto a row already `entra_obo` → 400. `_seed_model_def` (`:81-102`) takes neither field, so the PUT cases need it extended or the state set afterwards.

---

<details>
<summary><b>Worth fixing in this PR — 12 smaller items</b> (click to expand)</summary>

**7. Synthetic rows surface as a phantom MCP connection.** The listing filter (`mcp_oauth.py:4064`) keys on `obo_server_names`, so `__model_obo__:<aud>` rows have no `mcp_servers` row and pass straight through — rendering as a connection with a Disconnect button. On DELETE, the `oauth_obo` 409 backstop at `:4225` requires `server_row is not None`, so it's skipped and the row is deleted with a `token_revoked` audit — which the next model turn silently undoes by re-minting. Fix: a shared `_SYNTHETIC_TOKEN_PREFIXES` constant, filtered in the listing and 409'd in the revoke handler. Put the 409 **right after `server_name` is parsed** (`:4194-4196`), before the 404 existence check at `:4215` — otherwise 404-vs-409 becomes an oracle for whether a minted token exists for a given audience. The existing 409 body is worded for sign-in-passthrough servers; a model-mint row wants its own sentence.

**8. `entra_app` ignores `obo_grant_profile`.** `mint_app_access_token` hard-codes `scope=f"{audience}/.default"` with no profile branch (`:2963-2973`), while `mint_obo_access_token` dispatches via `_OBO_MINT_LEGS.get(profile)` at `:2746`. On `obo_grant_profile="rfc8693"` every `entra_app` alias sends a shape the IdP won't understand. Note there's no rfc8693 client-credentials leg to dispatch to — both entries in `_OBO_MINT_LEGS` are refresh-token grants — so "shape it per profile" means writing a new Keycloak leg, which is a feature, not a fix. Refuse with an explicit log when the profile isn't `entra`, and ideally reject at write time too (`console/server.py:11698-11706` can already read `app.state.oidc_config`).

**9. Every mint builds a throwaway `httpx.AsyncClient`** (`:2818-2820`, `:2960-2962`) — `app_state.obo_http_client` is set only under `tests/` and `scripts/obo-e2e/`, so production always takes the transient and pays DNS + a full TLS handshake. Don't reuse `app_state.mcp_oauth_http_client`: it's created on the lifespan loop (`:4488`) while the mints run on the mcp loop, and your own comment at `:2582-2590` explains why that collides. Create it in `MCPClientManager.start()` and close it in `shutdown()`.

**10. No in-process memo over the DB cache row.** Every warm turn still costs a cross-thread hop plus a `mcp_user_tokens` SELECT and a Fernet decrypt (`:2755-2757`). Put the memo inside `mint_obo_access_token` ahead of that read — everything there runs single-threaded on the mcp loop, so no lock is needed and the memo shares one invalidation point with the row. The freshness predicate must carry the same audience-equality check as `_is_fresh_obo_cache_row`, or an operator's audience narrowing stops taking effect. Land it *after* the deprovisioning purge (item 3), or the revocation window silently grows. (`tests/test_model_provider_obo.py:11` already describes this cache as if it exists — worth correcting either way.)

**11. Two lock-key families grow unbounded, and one prune is a no-op today.** `_refresh_lock_for` inserts and never evicts; `_drop_refresh_lock` is unreachable from this path. Separately, `_read_obo_credential` is called with the literal `"__model_obo__"` (`:2784-2786`) rather than the real cache key — so the prune inside `_no_token_result` targets a key that's never created. **Careful here:** passing the real key routes the missing-credential branch through `_no_token_result`, which calls `_drop_refresh_lock` *and* `_clear_refresh_backoff` — dropping the lock the mint is holding and clearing the cooldown item 2 just armed. It needs a `prune_on_missing=False` seam. And don't drop the `__obo__:<issuer>` lock — the MCP path shares that key, and it's one entry per user, not the growth term.

**12. `mint_app_access_token` duplicates the scaffolding.** `:2960-2962` is byte-identical to `:2818-2820`; the read-and-serve block appears four times in the new code (`:2755`, `:2776`, `:2928`, `:2943`). Extracting `_serve_fresh_mint_cache` and `_enter_mint_client` **first** means the cooldown, the memo and the long-lived client each get written once instead of twice. Don't fold the classified path's copy in — it carries a force_refresh token-identity comparison (`:2536-2538`) the model copies don't.

**13. `scopes` has no caller** (`:2715`, `:2752`) — `effective_scopes` is unconditionally `""` and the comment above advertises an rfc8693 restriction this path can't apply. Safe to drop. **Keep `force_refresh` though** — it's also caller-less today but it's the seam a 401-retry would need.

**14. `auth_mode` from `config.toml` is never validated** (`model_registry.py:529-530`) — no membership check, no cross-field check, so a typo silently falls back to the shared static key. The DB-row loader at `:455-456` has the same gap for a row written before the console validation existed. A `_normalize_auth_mode(alias, mode, audience)` helper called from both covers it in one edit. Note `_MODEL_AUTH_MODES` lives in `console/server.py:11258` and core mustn't import console — the house pattern (`_VALID_API_SURFACES`) is core-owns-it, console-mirrors-with-a-comment. Don't gate the placeholder on `obo_audience` in `get_client`: `create_client` converts an empty key to `None` so the SDK falls back to `ANTHROPIC_API_KEY`/`OPENAI_API_KEY`, which might *work* and mask the misconfiguration. The loud failure belongs at load time.

**15. `reload()` keeps the stale client when only `auth_mode` changes** (`model_registry.py:297-306`) — `client_key` derives from `auth_mode` at `:163-165` but the eviction condition doesn't compare it. Narrow window (needs `api_key` byte-identical and empty), but a one-line fix. Update the enumeration comment at `:292-296` too, and copy `test_reload_drops_client_when_base_url_changes` (`tests/test_model_registry.py:987`) for the regression.

**16. A failed mint silently downgrades to the shared static key** (`session.py:6039-6053`) with no fail-closed option. On a deployment that keeps both a shared key and `entra_obo` — precisely what your tests exercise (`api_key="static-fallback"`, `tests/test_model_provider_obo.py:553`) — IdP-side revocation doesn't deny model access and the gateway sees the shared identity. This reads as a deliberate contract, and it's a reasonable default; the ask is that it become an explicit operator-visible choice rather than the only behaviour. **Important:** a plain `raise` from `_model_obo_token` is *not* fail-closed — it's caught at `session.py:5386` and routed into `registry.fallback`, potentially a `static` alias with a shared key, which is the exact downgrade this would exist to stop. The refusal has to be absent from `provider.retryable_error_names` and re-raised before the fallback loop.

**17. The sub-agent lane resolves the token inside its retry loop** (`session.py:16133-16147`) — up to 4 resolves per turn, where the main path hoists it with the comment "Resolved once, outside the retry loop." (`:5497`). Straight hoist. (The "re-resolve after an auth failure" alternative doesn't exist yet — `_model_obo_token` has no `force_refresh` seam and neither does the sync bridge, so that's a bigger change than the hoist it'd replace.)

**18. `obo_audience` has no length cap** (`console/server.py:11897-11898`), while `alias`/`model` are truncated in the same handler and the MCP twin uses `_clean_oauth_text(max_length=2048)`. It becomes part of the `mcp_user_tokens` PK, so an oversized value overflows the Postgres btree limit, `_persist_obo_cache_row` raises, the `except Exception` at `:2850-2858` swallows it, and every turn does a full IdP round-trip. Two notes: `_clean_oauth_text` returns `str | None` and the column is NOT NULL, so coerce `or ""`; and it does *not* strip control characters despite the name — if you want that, add it inside the helper and the MCP path benefits too. The cap also doesn't bound the *runtime* value, because `_resolve_env_vars` runs on `obo_audience` at `model_registry.py:456/530` — a short `${VAR}` can expand past the limit at read time, per node. Was env-templating the audience intentional?

</details>

<details>
<summary><b>Smaller still — 4 items</b> (click to expand)</summary>

**19. The sync bridges abandon the mint future on timeout** (`mcp_client.py:1015-1017`, `:1043-1045`) — `concurrent.futures.TimeoutError` is swallowed by a bare `except` and the future is dropped, leaving a queued coroutine that still completes a full mint nobody reads. `call_tool_sync` (`:6837-6839`), which the docstring says this mirrors, does `future.cancel()`. Same fix here.

**20. The MCP freshness sweep's drive set now includes every model-mint row.** `list_mcp_user_token_reconcile_targets` runs an unfiltered full-table SELECT and its docstring states the invariant this PR breaks: "Only `oauth_user` servers write these rows." Prefer the `IN (<oauth_user names>)` form over `NOT LIKE` — `_` is a LIKE wildcard so the prefix needs `ESCAPE`, and Postgres `LIKE` is case-sensitive while SQLite's is ASCII-case-insensitive. **Three mirrored sites, not two**: `_postgresql.py:4820-4836`, `_sqlite.py:4974-4988`, and `_protocol.py:2250-2258`, which carries the same stale sentence.

**21. Parity site missed: `_protocol.py:2403-2420 create_model_definition`** was never given `auth_mode`/`obo_audience`, although both implementations have them. mypy doesn't catch it because console handlers hold `storage` as `Any`. Worth folding into whichever storage-layer fix lands first.

**22. Two consumers don't carry the new fields.** `server.py:internal_model_status:4194-4206` doesn't project them, so the console's synthetic rows for `config.toml`-sourced aliases (`console/server.py:11559-11577`) never carry them and the admin hint never fires for a TOML alias — its edit modal reads `undefined || "static"`. Separately, `sdk/typescript/openapi-console.json` is a checked-in generated artifact with zero `auth_mode` occurrences (no test asserts its freshness). And there's no `docs/*.md` or changelog entry yet — `docs/mcp-oauth.md`, `docs/oidc.md` and `docs/settings.md` are the natural homes.

**23. Migration 068 style (optional, safe to decline).** `052_model_reasoning_persistence.py` uses `batch_alter_table` on this same table, but 11 other migrations — including the two most recent column-adds — use the bare form, native `DROP COLUMN` works on any supported runtime, and your downgrade test passes. Consistency point only.

</details>

---

## Suggested landing order

Most of these are independent, but a few will fight each other if taken out of order:

1. **Helper extraction (12) first** — it's the carrier for the cooldown, the memo and the long-lived HTTP client. `mcp_oauth.py:2681-3000` is touched by seven fixes; land them as one ordered series (12 → 2 → 11 → 13 → 8 → 9 → 10), not seven parallel patches.
2. **Cooldown (2) before the lock-key change (11)** — reversed, the key change clears the cooldown you just armed.
3. **Mint reachability (1) before fail-closed (16)** — making the mint fail closed while it's unreachable on two of four hosts turns a silent 401 into a hard refusal for every CLI and console-coordinator turn.
4. **Mint reachability (1) before lane threading (5)** — each new lane inherits whatever reference (1) settles on.
5. **Unlink purge (3) before the memo (10)** — the memo must not outlive the purge.
6. **Admin 400 tests (6) before the permission gate (4)** — otherwise the new tests start returning 403 instead of the 400 they assert.
7. **Rename before threading**, if you take the naming item — same parameter, same doc paragraphs.

In `console/server.py:11690-11930`, take the cap (18) then the gate (4) — the gate's "did the effective value change" test wants the cleaned value in hand.

## Naming

`obo_api_key`, `_model_obo_token` and five doc sites (`model_registry.py:51-57`, `_schema.py:843-845`, `console_schemas.py:1014-1015`, migration 068's docstring, and `_model_obo_token`'s own) all describe a per-user OBO credential and a two-value enum. The enum has three values, and `session.py:6021` feeds an `entra_app` machine-identity token through that same "per-user" parameter. Anyone auditing gateway logs for attribution will read "per-user" and be wrong. Worth a rename to something mode-neutral — note `auth_token` collides with the OIDC/session vocabulary elsewhere in the tree, so something like `backend_auth_token` reads better at the binding site.

## Open questions for you

These we genuinely couldn't settle by reading, and your answers change the fixes:

1. **Are the judge / output-guard / perception lanes meant to carry the caller's identity, or the app's?** (See item 5 — the perception memo has no user component in its key.)
2. **Is the console's `MCPClientManager` deliberately app-state-less?** It decides whether item 1's fix should also wire the console.
3. **Was `_resolve_env_vars` on `obo_audience` intentional?** It makes the audience per-node templatable — feature or copied pattern?
4. **Does anything provision model definitions with a service token?** `allow_service_bypass=False` in item 4 would break it, and we found no in-tree caller besides the browser and tests.
5. **Does `with_options(api_key=…)` actually reuse the connection pool** on the pinned SDK versions? Both call sites assert it does. If it deep-copies the transport, every OBO turn opens a fresh TLS connection to the gateway — a much larger per-turn cost than anything in the performance items above. A one-line probe would settle it, and we'll run it if you haven't.

Again — genuinely good contribution, and the reuse-don't-fork instinct throughout is the right one. Most of the list above is reachability and failure-path work rather than anything wrong with the design.

metaclassing:
This was obviously generated through great amounts of divided attention and distraction, ping me on discord after work lol

# PR review response

Thanks for the thorough review and especially for tracing the interactions
between the proposed fixes. We addressed the findings as follows.

## Before merge

### 1. Mint reachability

Resolved. `ChatSession` now retains an ungated model-mint client separately
from the persona-gated MCP tool surface, so disabling MCP tools no longer
disables model authentication. A manager is also created when dynamic model
auth requires one even if there are no MCP servers.

Node, CLI, and console-coordinator hosts now wire storage and app state into
that manager. Here `app state` is the host dependency container, not an Entra
app identity. Runtime model reloads can create the manager when an alias first
switches to dynamic auth, and existing sessions receive the new reference.
Missing mint infrastructure emits a warning with the alias, mode, audience,
and whether a static fallback exists.

The console's missing app-state wiring was not intentional. We chose host
parity and wired it; this also makes the coordinator's existing per-user OAuth
pool machinery functional instead of leaving a partially initialized manager.

### 2. Mint failure cooldown

Resolved for both delegated and app mints. A fresh memo/DB token is served
before checking cooldown; the cooldown is checked before either lock; missing
credentials, grant failures, and malformed successful responses arm it; and
cache/mint success clears it.

The backoff key is the synthetic per-audience cache key. It never uses the
shared `__obo__:<issuer>` credential key, preserving the invariant that a
broken model audience cannot suppress the user's MCP OBO dispatches.

### 3. Identity deprovisioning

Resolved. Identity unlink now enumerates the user's token metadata and purges
only `__model_obo__:` rows in addition to the existing MCP OBO rows.
`__model_app__:` remains under the shared `__app__` principal and is not
user-deprovisioned.

The console also invalidates its coordinator memo and best-effort fans a
service-authenticated invalidation request to every registered node, closing
the process-local memo window on reachable hosts as well as deleting the
authoritative shared rows. The app-token revocation property is now explicit
in the docs: revoking the app credential prevents new mints, while an already
minted app bearer remains usable until its recorded expiry.

### 4. Audience constraint and permission boundary

Resolved with an exact operator allow-list:
`model.auth_audience_allowlist`. There is no host match or wildcard behavior.
Dynamic-auth writes validate the effective post-merge values and require
`admin.mcp` only when the mode or audience actually changes, so the admin
form's unchanged fields do not gate unrelated edits.

We chose `allow_service_bypass=False`. We found no in-tree service-token model
provisioner, and treating this as a capability-escalation boundary is safer.
We also gate a `base_url` change while an alias is dynamically authenticated;
otherwise an `admin.models` caller could retain the approved audience but
redirect its bearer to another host.

### 5. Model lanes

`ModelLane` now carries an optional backend-credential resolver, and
`model_turn` resolves it once before the drain/retry loop. Utility, judge,
output-guard, and perception lanes use that resolver. The main stream and
sub-agent keep their existing retry owners but resolve/bind the same
mode-neutral credential once outside those retry loops.

Judge and output-guard inherit the session's effective user: their verdicts
evaluate that user's turn and must preserve downstream attribution,
authorization, and accounting. Perception does the same after changing its
memo key to `(principal_id, alias, content_hash)`, so a description authorized
under one user's OBO grant cannot be served to another user. `entra_app`
remains available only when the model definition explicitly selects it; a
missing user or failed OBO mint never switches grant modes.

Eval and optimizer remain explicitly registry-less/raw-client development
tools, so dynamic model authentication is documented as unsupported there
rather than adding a token argument that cannot work.

The regression tests are behavioral: a configured `entra_app` lane binds the
minted value and never calls the provider with
`backend-auth-placeholder-unused`; session-built judge and output-guard lanes
hold the delegated resolver and mint with the effective user's ID rather than
the app grant.

### 6. Load-bearing tests

Added direct tests for the SDK `with_options(api_key=...)` binding using a
realistic mock client and a non-draining provider path. Added the four requested
admin validation cases:

- invalid `auth_mode`;
- dynamic mode without an audience;
- changing a row with no stored audience to `entra_obo`; and
- clearing the audience on an existing `entra_obo` row.

The suite also covers the new permission, cooldown, principal-partitioned
perception memo, profile, reload, fail-closed, deprovisioning, lane, restored
watch-owner, and synthetic-row behaviors.

## Additional findings

### 7. Synthetic rows in the MCP connection UI

Resolved with shared synthetic-prefix constants. Synthetic model cache rows are
filtered from the connection list, and revoke returns `409` immediately after
parsing the name, before the existence lookup. The response explains that the
row is a managed model-token cache rather than a revocable MCP consent.

### 8. `entra_app` grant profile

Resolved. Runtime minting refuses non-`entra` profiles with an explicit log,
and the admin write path rejects that configuration up front. We did not add
an RFC 8693 client-credentials leg under the guise of a bug fix.

### 9. Mint HTTP client lifetime

Resolved. `MCPClientManager` creates the model-auth `AsyncClient` on its own
event loop, attaches it to app state, reuses it for both mint types, and closes
it during manager shutdown. It does not reuse the lifespan-loop OAuth client.

### 10. In-process memo

Resolved. Both model mint paths check an MCP-loop-owned memo before the
encrypted DB row. The memo uses the same freshness and exact-audience predicate
as the DB cache. Fresh mints populate both; identity unlink deletes delegated
rows, invalidates the console memo, and requests invalidation from each
registered node.

### 11. Lock-key growth and missing-credential cleanup

Resolved. Synthetic per-audience locks are pruned after the final local waiter.
`_read_obo_credential` now has a `prune_on_missing=False` seam for model mints,
so the missing-credential path cannot drop a held synthetic lock or clear the
cooldown just armed. The shared `__obo__:<issuer>` credential lock is retained
because the classified MCP path also owns it.

### 12. Duplicated mint scaffolding

Resolved. The two paths share cache serving/memoization and mint-client context
helpers. The classified MCP path remains separate because of its
`force_refresh` token-identity semantics.

### 13. Unused `scopes`

Removed from `mint_obo_access_token`. `force_refresh` remains as the future
401-retry seam.

### 14. Core auth-mode validation

Resolved with a core-owned `MODEL_AUTH_MODES` set and normalization helper used
for both DB and TOML definitions. Unknown modes, missing dynamic audiences,
control characters, and audiences over the runtime cap now fail during
registry loading rather than silently falling back or consulting provider SDK
environment keys.

### 15. Stale client on auth-mode reload

Resolved. An auth-mode change participates in client eviction, with a regression
test modeled on the existing base-URL reload test.

### 16. Fail-closed policy

Added the operator-visible `model.auth_fail_closed` setting. When a real static
key is explicitly configured, the historical mint-failure fallback remains
available unless this setting is enabled. Two cases always fail closed
regardless of the setting: an `entra_obo` call has no effective user, or a
dynamic alias has no real static key. This prevents the SDK-construction
placeholder from producing an opaque 401.

`BackendAuthUnavailableError` is re-raised before provider retries and before
the registry fallback loop, so a refusal cannot downgrade to a shared fallback
alias. There is no OBO-to-client-credentials fallback; `entra_app` is used only
when explicitly configured on the selected alias.

### 17. Sub-agent retry loop

Resolved. The sub-agent token/resolver is now hoisted outside its retry loop,
matching the main lane.

### 18. Audience cap and environment expansion

Resolved. Admin input is control-character-cleaned and capped at 2048
characters, and core loading rejects controls and enforces the same cap. We
removed environment expansion from `obo_audience`: it was copied behavior
rather than an intended feature, and a security allow-list decision must not
vary or expand per node.

### 19. Timed-out mint futures

Resolved. Both sync bridges cancel their future on timeout before returning.

### 20. OAuth-user freshness sweep

Resolved in PostgreSQL and SQLite using an authoritative join to
`mcp_servers` with `auth_type = 'oauth_user'`. The storage protocol invariant
was updated, and tests verify synthetic model rows do not enter the sweep.

### 21. Storage protocol parity

Resolved. `StorageBackend.create_model_definition` now declares `auth_mode` and
`obo_audience`.

### 22. Status, generated artifact, and docs

Resolved. Node model status and console-synthesized rows carry both auth
fields. The checked-in console OpenAPI artifact was regenerated. The feature
and its security/lifecycle properties are documented in `docs/settings.md`,
`docs/oidc.md`, `docs/mcp-oauth.md`, and `CHANGELOG.md`.

### 23. Migration style

No change. We kept migration 068's bare additive/drop-column form because it
matches the more recent column-add migrations and the tested upgrade,
downgrade, and downgrade/upgrade round trip pass on the supported SQLite path.

## Naming

Renamed the mode-neutral model-call parameter and helper to
`backend_auth_token` and `_model_backend_auth_token`. Related schema,
migration, registry, and call-site wording now distinguishes delegated
`entra_obo` identity from shared `entra_app` identity. The lower-level
`mint_model_obo_token_sync` name remains specific because that function really
does mint only the delegated OBO variant.

## Open-question decisions

- Judge and output-guard carry the effective user identity.
- Perception carries the same identity and partitions its memo by principal.
- Console manager app-state wiring is enabled for host parity.
- `obo_audience` environment expansion is removed.
- Service tokens do not bypass `admin.mcp` for dynamic-auth capability changes.
- A probe against OpenAI SDK 2.48.0 and Anthropic SDK 0.120.0 confirmed that
  `with_options(api_key=...)` preserves the exact underlying `_client` object
  in both SDKs, so it reuses the connection pool.

## Follow-up identity trace

While resolving the lane-identity question, we found an existing Ringdown
restore gap outside the original review: an evicted watch was rehydrated with
`user_id=""` even though its persisted workstream has an owner. The restore now
reads that owner before constructing the unattended, auto-approved session and
passes it through to `ChatSession`. A missing owner is a permanent refusal; a
storage lookup failure is transient and retried. The server's startup
`--resume` path uses the same owner resolution.

This keeps transport identity separate from execution identity: the scheduler
may use a service credential to reach a node, but it already sends
`created_by` as the workstream user, and subsequent schedules, wakes, judges,
guards, perception calls, utility calls, and sub-agents continue under that
principal.

PR #938, opened 2026-07-31
feat(tool-search): surface MCP discovery failures & honest result counts

metaclassing:
TL;DR This adds some improved user visibility as to what is going on during complex obo tool discovery calls, where users may or may not have access to certain tool visibility, rather than simply returning no useful context.

---

Tool discovery for a pool-backed (oauth_user/oauth_obo) MCP server that is down, unauthorized, or 5xx-ing was invisible in chat: the server contributed zero tools to the catalog, so tool_search returned the misleading "No matching tools found. Try a different search query." — indistinguishable from a genuine no-match. Partial results were also silently truncated to max_results with no signal that more matched.

This makes both cases explicit, without touching the token/OBO machinery:

- tool_search.ToolSearchManager
  - search() now ranks the whole deferred corpus and records the full (pre-slice) match count, so format_search_results reports honest truncation ("Showing the top N of M — narrow your query").
  - Optional status_provider callback -> {server: get_server_status()} lets results flag servers that are actually failing (open circuit breaker, recorded error, or recorded discovery failure). A down server is named with its reason instead of masquerading as "no such tool". Deliberately conservative: a merely un-primed server (connected=False, no error) is NOT flagged, to avoid crying wolf. A provider that raises is treated as "no info" and never breaks search.

- mcp_client.MCPClientManager
  - New _pool_discovery_error map: the previously-swallowed pool prime/connect discovery failure (the exact 500-at-discovery case) is now recorded and cleared on the next successful connect, and exposed as get_server_status(...)["discovery_error"]. Additive only.

- session.ChatSession wires get_all_server_status(user_id) into both ToolSearchManager constructions via _mcp_status_snapshot.

Scoped to merge cleanly alongside feature/model-provider-obo: all edits land in tool_search.py (untouched by that PR) and in disjoint regions of mcp_client.py / session.py; mcp_oauth.py and the system-turn/admin-UI surfaces are deliberately left untouched.

Tests: extends tests/test_tool_search.py (truncation counts, unavailable-server advisory across error/circuit/discovery_error, healthy & un-primed not flagged, no-provider legacy behaviour, provider-error resilience, _status_reason).


eous:
Merged — thanks! Reviewed against the head revision: the truncation count is exact (BM25 only returns positive-score matches, and the 50-doc rerank pool keeps result order identical to before), the discovery-error lifecycle (record at prime, clear on connect/remove/reconcile-flip) all checked out, and the earlier review points were addressed cleanly with tests.

One thing we'll harden separately (#941): `_pool_discovery_error` is name-keyed while pools are per-(user, server), so one user's prime failure can surface the unavailable advisory to a user whose own pool is warm. Advisory-only and self-healing, so it didn't need to hold up the merge.

Issue #955, opened 2026-08-03
Model auth on rfc8693 (Keycloak): add a dedicated rfc8693_obo mode + per-alias scopes — entra_obo is overloaded and cannot mint


eous:
Live-verified with `scripts/obo-e2e/keycloak_e2e.sh` (Keycloak 26.3, standard
token exchange): a model alias with `auth_mode=entra_obo` on an
`obo_grant_profile="rfc8693"` deployment cannot mint. (Reading note:
`entra_obo` is **today's only user-context mode** — it is currently overloaded
to serve BOTH grant profiles, which is why an Entra-named mode appears in an
rfc8693 repro at all. Removing that overload with a dedicated `rfc8693_obo`
mode is the Direction below.) Both legs run (refresh → exchange, 2 IdP calls
observed), and the exchange is refused:

    HTTP 400: error=invalid_request
    error_description=Requested audience not available: mcp-a

The MCP `oauth_obo` path minting for the **same audience with the same captured
credential succeeds** in the same run (checks E1–E7 all VERIFIED).

## Mechanism

`_obo_mint_rfc8693`'s exchange leg sends `scope` only when the caller supplies
one. The MCP caller has a per-row scope source (`mcp_servers.oauth_scopes`);
the model caller has none — `model_definitions` carries `obo_audience` but no
scopes column, and #898 removed the dead `scopes` parameter from
`mint_obo_access_token` (correctly; nothing could supply it). The combination
was structurally unmintable from the day it shipped.

Attaching the audience mapper as a **default** client scope on the app client
was tried and does **not** clear the refusal on KC 26 standard token exchange —
do not document that as a workaround without re-verifying.

## Direction (2026-08-03)

Close the gap with a **dedicated `rfc8693_obo` auth mode** plus a per-alias
`obo_scopes` column, rather than overloading the `entra_*` modes. (Working
name: `rfc8693_obo` parallels `entra_obo` and leaves `rfc8693_app` open for a
future app-credential variant instead of squatting the bare profile name;
final spelling is the implementing session's call.)

- Operator clarity: an operator on Keycloak/generic rfc8693 picks the mode that
  names their IdP dialect instead of an Entra-named mode that half-works.
- Code clarity: mode→mint-leg dispatch becomes explicit instead of
  profile-conditional behavior inside an Entra-named path.
- It makes **write-time profile pairing enforceable**: with distinct modes,
[...]

Issue #954, opened 2026-08-03
Model backend auth is gated on admin.mcp; needs its own permission and a multi-sender identity ruling


eous:
Configuring a model's backend auth is gated on `admin.mcp` today. That was
chosen in #898 by analogy — "writing `oauth_audience` on an MCP server is the
same capability" — but the field lives on `model_definitions`, which is
administered under `admin.models`. The practical effect is that an operator who
should only manage models must be granted full MCP-server administration,
including OAuth client secrets, to direct model credentials. Wrong subsystem,
and a much wider grant than the task needs. `admin.models` alone is not
sufficient either, which is why the analogy got reached for in the first place.

## The hazard a second gate exists to cover

`model.auth_audience_allowlist` bounds which **audience** may be redeemed
(write needs `admin.settings`). It does not bound where the minted token is
**sent** (`base_url`) or how the request is shaped (`capabilities` —
`server_compat.api_surface` selects the provider factory, `extra_body` is
merged into provider requests). Without a second gate, an `admin.models`
holder could point an approved audience at a host they control and collect
every user's minted gateway token. The allow-list constrains the credential's
*subject*, never its *destination*.

## The gate as it exists now (post-#950 — orient here, the surface moved)

The #950 branch replaced the original enumerated-field gate with **default-deny
field classification**, and any new permission must gate the same boundary:

- `_derive_auth_gate(existing, updates) → ModelAuthGateDecision`
  (`turnstone/console/server.py`, pure function, unit-pinned invariants) is the
  single derivation both write twins consume. **A permission split swaps the
  scope at the decision's consumers — one family of sites — not at a field
  list.** `MODEL_AUTH_NEUTRAL_FIELDS` enumerates the provably auth-neutral
  columns; every other change on a row that is-or-becomes dynamic requires the
  gate + `_validate_dynamic_model_auth` (row tier = audience allow-list always;
  posture tier on `posture_event` = pair change or enable-arming).
- **Disarm is deliberately below the gate**: `enabled` true→false as the only
  gated change needs plain `admin.models` and skips validation (de-escalation
  must never be blocked — pinned). The split question is therefore only about
  *escalation/aiming*; do not regress this.
- **Calibrate is a guarded exception**: it writes the gated `capabilities`
  column under `admin.models`, protected by an *enforced* confinement
  invariant (`calibration_confinement_violations()` → 500 + CAS persist), not
  by the gate. A new permission must not accidentally capture it.
[...]

PR #961, opened 2026-08-04
feat(models): default-deny governance and admin UI for per-alias backend auth

### LINKED ISSUE #950: Model auth admin UI does not surface the allow-list, capability, or grant-profile gates
The model-definition admin modal gained an Auth mode / OBO audience pair in
#898, but it implements the server contract as it stood *before* the review
hardening. Three server-side rejections were added afterwards and none of them
reached the UI, so each is discoverable only by submitting the form and reading
a 400 or 403.

## The gaps

**1. The audience allow-list is invisible.** `admin_create_model_definition` /
`admin_update_model_definition` require `obo_audience` to be an exact member of
`model.auth_audience_allowlist` (`console/server.py:11459`), which defaults to
empty — deny-all. Nothing in the modal says the value must be pre-registered,
and there is no way to see what is registered. Out of the box every save of a
dynamic alias fails until the operator finds an unrelated settings field.

The placeholder makes it worse: `index.html` offers
`https://your-resource.example.com` while the allow-list's own help text says
"Use literal values such as `api://<application-id>`", so following the
placeholder produces a value that was almost certainly never registered.

*Fix:* source the field from the allow-list — a `select` or `datalist` over
`model.auth_audience_allowlist` instead of free text. A deny-all default then
renders as an empty picker with a "no audiences registered" hint pointing at the
setting, and the placeholder mismatch disappears with the free-text box.

**2. No `admin.mcp` signal.** Changing mode or audience requires `admin.mcp`,
not just `admin.models`, with no service bypass. An `admin.models`-only operator
gets editable fields and a 403 on save. Disable the pair and annotate it when the
caller lacks the capability.

**3. `entra_app` is offered where it cannot work.** The server rejects
`entra_app` unless `[oidc] obo_grant_profile='entra'`
(`console/server.py:11475`). The dropdown offers it unconditionally, so on a
non-Entra deployment selecting it is a guaranteed 400. Hide or disable the
option with a reason when the deployment profile is not `entra`.

**4. The `base_url` gate is unsignalled.** Changing `base_url` while an alias is
dynamically authenticated is rejected. Same discover-by-400 shape.

## Smaller

- The list-view hint chips push bare `obo` / `app` strings
  (`admin.js:7162-7165`). Plain-language wording, consistent with how the other
  override hints read.
- The OBO audience input stays enabled when Auth mode is `static`. Harmless —
  the server evaluates whether the effective value actually changed — but it
  invites confusion.
- `_watch_restore_owner` is called with a fresh `_get_storage()` at both sites
  (`server.py:5824`, `:5935`) rather than a handle already in scope.

## Not in scope here

`entra_app` under a non-Entra grant profile needs a client-credentials leg that
does not exist yet; that is a feature, and #898 was right to refuse rather than
write one. Eval and optimizer lanes remain registry-less and are documented as
unsupported for dynamic auth.

eous:
Follow-up to #898's per-alias backend auth (`entra_obo` / `entra_app`): the console write path gains default-deny governance, the admin shelf gets full backend-auth support, and the registry/session rebind machinery is hardened for config changes landing under live sessions. Single commit; the commit body carries the full change inventory.

## Review guide

**Where the newest logic lives** (most valuable reviewer attention):
- `turnstone/core/session.py` — per-send rebind on registry generation change: atomic `resolve_binding()` commits, fallback-first handling of removed/unconstructable aliases, judge/limiter resets gated on the binding actually changing.
- `turnstone/core/model_registry.py` — the `reload()` chokepoint enforces the dynamic-auth encryption-key requirement and bumps the generation before the map swap (ordering argument in comments at both sites).
- `turnstone/console/server.py` calibrate path — confinement invariant plus a compare-and-swap persist through a new `expected_capabilities` storage parameter.

**What is most heavily verified** (six adversarial review rounds ended with zero findings here):
- The write gate: `MODEL_AUTH_NEUTRAL_FIELDS` default-deny classification, `_derive_auth_gate` (pure function, unit-pinned exclusivity invariants), the two-tier validator, the pure-disable de-escalation carve-out, and the live-schema classification test that forces every future `model_definitions` column to be classified.

**Deliberate behaviors reviewers may trip on** (each pinned by a test named for it):
- Disabling a dynamic row needs only `admin.models` and skips validation — de-escalation is never blocked, even for rows with corrupt/skewed stored values.
- Enabling is a posture event even when the (mode, audience) pair is unchanged.
- The audience allow-list is write-time governance; stopping active minting is the disable lever (documented in `docs/oidc.md`).
- The permission helpers in `auth.js` have a deliberate 3-line parse fallback duplicated at the call-site shims — cache-skew survival, bound by lockstep comments in both files.

## Verification

- Suite: 10417 collected / 10407 passed / 10 skipped (99 tests added on this branch), plus a 71-scenario headless browser harness over the real admin shelf.
- Live rfc8693 token-exchange e2e: MCP legs verified end-to-end; the model-leg scope gap is tracked as #955 under a narrow known-gap signature.
- OpenAPI specs regenerated and freshness-tested; `mypy`/`ruff`/`prettier` clean.

## Related

Closes #950. Known follow-ups filed with orientation docs for cold pickup: #954 (dedicated model-auth permission and the multi-sender identity questions), #955 (dedicated `rfc8693_obo` mode + per-alias scopes).


eous:
Automated review feedback addressed in 78e92ebd: the seven duplicate-import-style flags are consolidated onto one from-style module import per file (module objects stay available for monkeypatching), and the JSON body read is now typed as Any until the isinstance narrow earns the dict annotation. The two code-scanning alerts are dismissed as false positives with reasoning attached: the flagged log line carries the missing-key remediation hint (config knob names, no secret value exists at that point), and the flagged URL check is a test assertion on a rebind result, not sanitization logic. Full suite re-run green: 10407 passed / 10 skipped.

Issue #983, opened 2026-08-06
Inline reasoning-tag scan consumes quoted tag vocabulary — content ABOUT the tags cannot round-trip


eous:
## Live incident (2026-08-05)

A deployed review exercise produced a report that discussed the think-tag splitter — and the report got mangled in transit: its own example content containing tag literals was rerouted into a reasoning block mid-stream. Earlier in the same run, the reviewing agent had already learned to hex-encode the tag strings (`3c7468696e6b3e` → `<think>`) and reconstruct them inside a python subprocess, because that was the only way its byte-level probes survived its own harness.

## Problem

The reasoning-tag vocabulary (`<think>`/`<reasoning>` + closes) is in-band, and some layer claims it on **every** lane posture:

- scan-on lane (passthrough server): `ThinkTagSplitter` / the drain's `split_inline_reasoning` consume quoted tags and reroute the following text to the reasoning lane;
- scan-off lane (`server_parses_reasoning`): the *server's* reasoning parser eats them before turnstone ever sees the stream.

So an assistant turn that merely QUOTES the vocabulary — a code review of the splitter, docs about reasoning models, a user pasting a transcript — cannot render or commit faithfully. Markdown code fences do not protect: the splitter has no fence awareness, and server-side parsers are fence-blind too. Now that agents review this codebase from inside the harness, meta-content is a real traffic class, not a curiosity.

## Candidate mitigations (turnstone's layer only)

1. **Anchor open-tag recognition to the start of the response** — real reasoning models emit `<think>` at position 0 (possibly after whitespace), before any visible content. Restricting OPEN-tag recognition to before-the-first-content-flush lets mid-response quoted tags pass through while preserving the actual use case. Needs a ruling on interleaved-thinking shapes (a mid-turn re-open after tool results) before committing.
2. **Fence-aware scan suppression** — don't scan inside markdown code fences. Cheap state machine, directly covers the quoting-in-reports case; heuristic rather than principled.

Both are splitter-behavior changes with parity implications, so 1.9-shaped. The current always-scan-everywhere behavior predates the 1.7 era and carries no recorded rationale — improve-and-name applies, not preserve-and-pin.

## Out of scope

The server-side parser layer cannot be fixed from here: on `server_parses_reasoning` lanes the tags are consumed upstream regardless. Worth a per-lane documentation note ("this backend's parser owns the tag vocabulary; quoted tags will not survive") rather than code.

Issue #982, opened 2026-08-06
Task agents: show used/total context while running


eous:
## Problem

A long-running task agent gives the operator no view of its context consumption. Watching a 262-step agent run today there was no way to answer "will it finish its synthesis before it eats its window" — which matters more than it looks, because the sub-agent loop has **no trajectory compaction**: with `tools.agent_max_turns` at its default (-1, unlimited), the de-facto terminator for a long run is context overflow, at which point the loop salvages the last assistant text and stops. Context used/total is effectively the progress bar for long runs, and it is invisible.

## The data is already in hand — this is surfacing, not accounting

- `_run_agent`'s `_api_call` already captures per-turn usage: it calls `_record_aux_usage(agent_result.usage, ...)` after every model turn. `usage.prompt_tokens` of the latest call IS the trajectory-as-billed — the honest "used" number.
- The denominator is already resolved: `run_task_agent` computes `agent_caps` for the agent's own lane, so `agent_caps.context_window` is correct even when the agent runs a different alias than the session.
- The emission channel has a precedent: `_paint_agent_step` → getattr-guarded `ui.on_agent_step(parent_call_id, item)` — web nests it under the task card; CLI and eval fixtures no-op automatically.

## Proposed shape

1. **Server hook**: after each `_api_call` return, emit the reading under the parent call_id — either a sibling hook (`on_agent_context(parent_call_id, used, window)`) or an extra field on the existing step item.
2. **Web**: a badge on the running task card — `41k / 128k` with a warning color from ~80% (approaching overflow-salvage).
3. **Refresh-resume**: the latest reading must ride the inflight snapshot so a reconnect mid-run re-paints the badge — live-event-only would vanish on refresh. (This is the bulk of the work, as with every live-card affordance.)
4. **Absent-usage lanes**: when a backend reports no usage, show nothing — no `?` badges. A char/4 estimator back-fill is possible later; silence is the right v1.
5. **CLI**: optional — append the reading to the existing `[task]` progress lines on decile changes.

## Follow-on (separate issue when wanted)

The same feed enables a budget-shaped stop. Today the only knob is `tools.agent_max_turns` (-1 or 1..200, model turns): there is nothing between 200 and unlimited, and no token-denominated bound at all. A "stop and synthesize at N% of window" rule reading this same number is the natural enforcement follow-up — visibility first.

[...]

Issue #981, opened 2026-08-05
commit-to-save window: a /history fetch between on_turn_committed and save_message renders the turn missing


eous:
Pre-existing (documented during #832's dataflow mapping, deliberately not widened by the fold): `on_turn_committed` clears the inflight buffers when the assistant turn enters the in-memory history, but the storage row lands at `save_message` a few statements later. `/history` is a REST fetch from storage, so a fresh connect landing in that window sees neither the row (storage doesn't have it yet) nor the snapshot (buffers already cleared) — the turn is invisible until the next event repaints.

Directions, undecided: reorder so the row lands before the buffer clear (couples the UI event to a storage write on the hot path), or retain the snapshot until the save confirms (a handoff latch in the UI base). The window is a few statements wide today; any change to the commit tail should re-measure rather than assume.

Issue #980, opened 2026-08-05
StreamChunk.is_first is produced by three adapters and consumed by nothing — delete it


eous:
Field audit during #832 (confirmed by an adversarial pass): `StreamChunk.is_first` is written by the OpenAI chat adapter, the Responses adapter (six sites), and the Anthropic adapter — and read by nothing in product code or tests (the only test references either produce it or assert the providers set it, pinning a contract no consumer holds).

Pre-1.7-era vestige. Delete the field, the producer writes, and the producer-side test assertions. One sweep, three adapters, no behavior.

PR #984, opened 2026-08-06
Fold the main streaming loop onto model_turn (#832)

### LINKED ISSUE #832: model_turn as the universal turn interface: stream the main loop through it, sever ChatSession from providers
## Goal state (does not gate 1.8)

After #827 and #831, every lane except the main streaming loop speaks Turn IR through `model_turn()`. This issue finishes the job: the main loop's send path becomes a `model_turn` call with a chunk callback, and `ChatSession` stops touching providers entirely.

The main loop is already funneled for this: exactly one `create_streaming` call site (`_try_stream`), fed by `_create_stream_with_retry` / `_try_fallback`, consumed by a single assembly method that already ends in the shared `_finalize_provider_blocks`.

## Shape

- `model_turn(lane, turns, *, tools=..., on_chunk=None) -> Turn` — with `on_chunk` absent, the drained transport of the previous phase; with it present, each normalized `StreamChunk` is surfaced to the caller as it arrives and the final assembled `Turn` is returned as usual.
- Chunk→Turn assembly moves under `model_turn`: one assembler for all lanes.
- Chunk→UI translation (on_status, early-paint `tool_pending`, cancellation generations, SSE) stays in `ChatSession` as the callback body — the harness keeps ρ and the event surface; it just stops owning the plant call.
- `_try_fallback` becomes lane-swap retry: a fallback alias is just a different `ModelLane`.

## Acceptance

- `ChatSession` imports no provider module and contains no provider-facing code; `model_turn` is the single plant-call surface in the codebase, and the provider Protocol's single method has one caller module.
- Main-loop behavior identical under the UI event stream: early-paint, mid-stream cancellation, usage accounting, and fallback retry all preserved.

## Related

- Depends on #831. Sequenced last deliberately: by the time the main loop moves, `model_turn` has been hardened under every other lane, including live Gemini judges (#827).

eous:
Advances #832. The interactive send path no longer owns the plant call: it makes one `model_turn` call per attempt with a chunk callback, and chunk→Turn assembly moves under `drain_stream` with every other lane. `ChatSession` keeps ρ and the event surface — chunk→UI translation, cancellation generations, commit, calibration.

## What changed

`model_turn` gains a streaming surface: `on_chunk` tees each normalized chunk to the caller upstream of the drain, so the callback sees exactly the sequence the assembler consumes, and the assembled Turn comes back as usual. `prepare_wire` lets a caller compose its own deterministic lowering — called with the serving lane, so a fallback folds system turns on its own capabilities rather than the primary's. The result carries `wire_msgs` for calibration and an accurate `producer`.

Deleted from the session: the hand-rolled chunk assembler, the stream-creation ladder, and the fallback walk's stream plumbing. `_try_fallback` becomes a lane swap. Cancellation converges — every model-call site now builds a fresh generation-scoped `_CancelRef`, closing the force-cancel hole where the old shared gen-0 ref read `aborted=False` for an orphaned generation.

Ladder arithmetic is unchanged: per-lane creation retries inside the fallback walk, wrapped by the mid-stream re-issue ladder, with no third multiplicative layer. Health still records success at the request-accepted instant and failure once per lane's whole ladder.

## What it does not do

The session no longer **calls** the provider Protocol on the plant path, and imports no provider module — but it still **holds** provider and client handles across ~30 sites, and reads three attributes off them (compaction's producer stamp and retry predicate, the fatal formatter's backend label). That is the remaining half of #832's acceptance, enumerated with file:line and a closure criterion in turnstonelabs/turnstone#979; that issue closing closes #832.

## Behavior changes worth a release note

- **Trailing citation footers now enter the committed turn.** Web-search source lists were an ephemeral info line that never survived a reload; they now fold into content on the same conditional rule the drain uses. Consequence accepted deliberately: web-controlled text reaches conversation storage, full-text search, later model context, and exports.
- **A stream that ends with no finish reason is a mid-stream death.** It used to commit its partial silently; it now re-issues through the mid-stream ladder and surfaces an honest error if it never recovers. Servers that legitimately omit terminal signals keep the existing `finish_reason_optional` shim.
- **The displayed stream and the committed turn now agree at reasoning boundaries.** Previously both dropped the splitter's held tail; the fold made the commit correct and left the display behind, so this closes the gap in the other direction too.
- Smaller ones: stored reasoning is segregated rather than interleaved for mixed-mode models; `producer` names the lane that actually served the turn, so a fallback-served turn stops wearing the primary's name; a superseded generation now exits silently as cancelled instead of propagating its exception class into the thread runner; and with debug enabled the request dump reprints after compact-and-retry, where the wire genuinely changed.

## Verification

Full suite green at 10,704 passed / 10 skipped, with strict mypy and ruff clean. A replay-parity harness pins the streaming phase against baselines captured from the pre-fold path: the same chunk script must produce the same UI event sequence and the same committed message, and every transform applied to a baseline restates in full the behavior change that justifies it. A mirror test asserts the displayed content and the committed content are identical across adversarial interleaves of content, reasoning, tool-call and info chunks. Golden wire payloads were untouched by the fold.

Six review rounds ran against the branch, each fix pinned by a test that was mutation-probed — reverting the fix must fail its pin — including the cancellation guards that only fire in a race window and had never been executed by the suite before.

## Follow-ups filed

- turnstonelabs/turnstone#979 — the session still holds provider/client handles; carries the full residual inventory that closes #832
- turnstonelabs/turnstone#980 — delete the unused `is_first` chunk field
- turnstonelabs/turnstone#981 — commit→save window in `/history`
- turnstonelabs/turnstone#982 — no used/total context readout while a task agent runs
- turnstonelabs/turnstone#983 — the inline reasoning-tag vocabulary is in-band, so content *about* those tags cannot round-trip

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -132,10 +132,12 @@ A user message flows through the system as follows:
      |                                  lane-swap fallback walk; per-lane ladder:
      |                                  up to 3 retries (4 total attempts), exponential backoff
      v
- the on_chunk consumer  ----------->  dispatch tokens to UI:
+ the on_chunk consumer  ----------->  display grid ONLY:
      |                                  on_reasoning_token() / on_content_token()
-     |                                  accumulate tool_calls from deltas
-     |                                  track finish_reason
+     |                                  tool-call deltas just flush the splitter
+     |                                    (assembly lives in drain_stream, inside
+     |                                    model_turn — the consumer never accumulates)
+     |                                  track finish_reason (citations-footer gate)
      |                                  _check_cancelled() per chunk (cooperative cancel)
      v
  finish_reason check:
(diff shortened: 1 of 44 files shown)


eous:
Static-analysis findings triaged. Two were real and are fixed in 76238320:

- `tests/test_cancel.py` had an `if True:` wrapper left behind when a patch block was mechanically converted to the armed-provider fake — a nesting level that manages nothing. Removed; the file previously carried a `nullcontext()` leftover of the same shape and now has neither.
- The parity runner's pre-fold branch called the streaming seam with two arguments. That is correct only against a tree whose signature still takes the wire list, so against this tree it reads as an arity error to a checker and to a reader. It now splats a named tuple, which states that the two-argument form belongs to the other world.

The remaining six are false positives, recorded here so they are not re-litigated:

- Four "unused import" hits in `model_turn.py` (`create_provider`, `TRAILING_INFO_SEPARATOR`, `folds_trailing_info`, `merge_usage`) are deliberate re-exports written with the explicit `X as X` idiom. `ChatSession` imports all four *from this module* so the provider package keeps a single import site — that is the point of the seam, and each is consumed in `session.py`.
- Two "except block handles BaseException" hits are intentional and already carry `# noqa: BLE001`. Both sit in harnesses whose job is to record *whatever* the seam raised: `GenerationCancelled` subclasses `BaseException` precisely so ordinary handlers cannot swallow it, so narrowing these to `Exception` would make the harness blind to the cancellation behavior it exists to observe.

PR #1045, opened 2026-08-20
feat(webui): allow specifying MCP OAuth 2.0 trusted pivate hosts

COM8:
## Problem Description

In a corporate network Turnstone and for example your local GitLab are running on the same private network. When you now try to configure OAuth for example for the GitLab MCP server, Turnstone blocks this since the GitLab domain resolves to a private network (e.g. `gitlab.example.com` -> `192.168.1.2`).

## Changes

* New optional `TURNSTONE_MCP_OAUTH_TRUSTED_PRIVATE_HOSTS` environment variable with a list of trusted local IPs.
* Frontend support for adding additional trusted hosts.

image

In the picture above, you see the settings page with the new `oauth_trusted_private_hosts` option. The one host with the `ENVIRONMENT` tag was set via an environment variable. The host with the `MANUAL` tag was added manually via the webif and can be removed again.

## TODOs

- [x] Test
- [x] Updated docs


COM8:
Switching back to draft since most of the functionality is now available upstream (besides the UI).

Issue #1066, opened 2026-08-29
UVX not found when used in the Docker container


BoguesUser:
Uv exist but uvx was not added to the container. 

I got it working on my end by changing line 11 in the docker file to `COPY --from=ghcr.io/astral-sh/uv:0.12.5 /uv /uvx /usr/local/bin/` and rebuilding the docker containers with `docker compose build --no-cache`

PR #1088, opened 2026-09-04
Adds the uvx command

BoguesUser:
Fixes issue #1066 where uvx would not be found when using mcp server that require it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ FROM python:3.14-slim
 LABEL org.opencontainers.image.title="turnstone" \
       org.opencontainers.image.description="Multi-node AI orchestration platform"
 
-COPY --from=ghcr.io/astral-sh/uv:0.12.9 /uv /uvs /usr/local/bin
+COPY --from=ghcr.io/astral-sh/uv:0.12.9 /uv /uvx /usr/local/bin/
 
 # Remove the slim image's man page exclusion so man-db has actual content
 RUN rm -f /etc/dpkg/dpkg.cfg.d/docker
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -8,7 +8,7 @@ FROM python:3.14-slim
 LABEL org.opencontainers.image.title="turnstone" \
       org.opencontainers.image.description="Multi-node AI orchestration platform"
 
-COPY --from=ghcr.io/astral-sh/uv:0.12.9 /uv /usr/local/bin/uv
+COPY --from=ghcr.io/astral-sh/uv:0.12.9 /uv /uvs /usr/local/bin
```
eous:
```suggestion
COPY --from=ghcr.io/astral-sh/uv:0.12.9 /uv /uvx /usr/local/bin/
```
Change `/uvs` to `/uvx`, and add a trailing slash to the destination so Docker treats it as a directory.

eous (commented):
Thanks for this! Adding `uvx` is the right call. The MCP registry docs already say that PyPI-based servers run through `uvx`, so the image should ship it.

Two small fixes are needed before this will build:

1. The binary is called `/uvx`, not `/uvs`. The uv image ships two binaries, `/uv` and `/uvx`. As written, the build fails because `/uvs` does not exist.
2. When `COPY` has more than one source, Docker requires the destination to be a directory that ends with a slash. So it needs to be `/usr/local/bin/` rather than `/usr/local/bin`.

One thing to know: CI does not build the Dockerfile, so the checks on this PR stay green even when the `COPY` line is broken. Please run `docker build .` locally after the change to confirm it builds.

There is a suggestion on the changed line below that you can apply with one click.

BoguesUser:
Sorry about the typos!
I made the edit in the browser so never noticed them. Thank you for merging the changes.

PR #1201, opened 2026-09-20
Fix reasoning replay for resumed Responses tool rounds

RedEyeNinja-BKK:
## Problem

On replay-sensitive OpenAI Responses tool rounds, Turnstone can reconstruct an in-round assistant
history item without replayable reasoning state. Providers that require reasoning replay then reject
the resumed request:

    HTTP 400: The `reasoning_text` in the thinking mode must be passed back to the API.

`_convert_messages` replays reasoning only from a stored native `reasoning` block, so a round whose
interrupted assistant turn recorded no reasoning is reconstructed as `function_call` +
`function_call_output` with **no reasoning item at all**. A strict-thinking upstream refuses that
history and the request fails instead of degrading gracefully.

The trigger is a compaction-shaped resume: the in-band summary is written by a non-thinking lane, so
the resumed history continues from an assistant turn (tool calls plus their results) with nothing to
replay.

## Reproducer

    user:      "Weather in Paris?"
    assistant: tool_calls: get_weather        # no reasoning recorded for this turn
    tool:      "18C, clear."                  # history ends here

Hermetic before/after:
`tests/test_provider_openai_responses_round_replay_repair.py::TestStrictUpstreamContract`.

## Change

The change synthesizes a valid Responses-native reasoning item **only for the affected current
round**, when the existing replay capability is enabled and no real stored reasoning is available:

    {"type": "reasoning",
     "id": "rs_roundrepair_<24 hex chars, digest of the answered call ids>",
     "summary": [{"type": "summary_text",   "text": "..."}],
     "content": [{"type": "reasoning_text", "text": "..."}]}

The text records that no reasoning was recorded for the turn; it does not invent reasoning. A
`reasoning_text` content part is carried because a summary-only item does not satisfy the contract.

Properties:

- **Real stored reasoning is preserved** — the pass only acts where no reasoning item exists.
- **No duplicate reasoning is introduced** — the added item cannot coexist with an already-present one.
- **Ordinary histories are unchanged** — nothing fires outside the resumed-round shape below.
- **Gate-off behaviour is unchanged** — gated on `replay_reasoning_to_model` (the flag `_build_kwargs`
  already receives, AND-gated with `caps.supports_reasoning_replay` in
  `model_turn.resolve_replay_reasoning_to_model`), so a lane that does not replay reasoning sees a
  byte-for-byte unchanged request.
- **Synthesized IDs are deterministic** — the id digests the *set* of answered call ids, so it is
  stable across retries and independent of history ordering.
- **The transformation is idempotent** — an own reasoning item is one of the guards, so re-running
  the pass is a no-op.
- **Pure transform** — returns a new list and never mutates its input.

## Scope of the repair

It fires only when **all** of these hold:

- the history ends on `function_call_output` — a tool-round continuation. "The conversation mentions
  tools somewhere" is not a trigger;
- the turn preceding that output carries at least one `function_call`;
- that turn already carries no `reasoning` item;
- every trailing output is answered by a call in that turn — an orphan, stale or reordered output
  fails closed rather than repairing the wrong turn;
- a user message precedes the turn, so historical assistant turns are never decorated.

Otherwise the `input` array comes back untouched.

## Compatibility

No wire-shape change outside the resumed-round case above; no new configuration; no new dependency;
no change to the chat wire, the Anthropic wire, or the storage format.

## Tests

- `tests/test_provider_openai_responses_round_replay_repair.py` (new, 30 cases): the repaired shape,
  and every neighbouring shape that must not change (plain turn, completed round, unanswered calls,
  mispaired/stale output, non-string call ids, no user message, empty input, output-only input),
  plus determinism, idempotence, non-mutation, schema and JSON-serializability.
- Three `tests/data/wire_payloads/openai_responses__*.json` goldens are updated. **This is part of
  the fix's regression contract, not a snapshot refresh.** Each of the three
  (`native_orphan`, `native_reasoning`, `trailing_orphan`) freezes a resumed tool round with replay
  enabled, so each previously encoded exactly the rejected shape. The fixtures change because the
  repaired Responses history now contains the required synthesized reasoning item — the diff in each
  is that added item and nothing else, and `toolcall_complete` / `mid_orphan` are unchanged. Each
  update is load-bearing: reverting it fails its corresponding test. Because the synthesized id is
  deterministic, the fixtures are reproducible.
- `CHANGELOG.md`: one entry under `[Unreleased] → Fixed`.

## Validation

- `ruff check` and `ruff format --check`: clean on the changed files. `mypy turnstone`: clean.
- Focused sweep across the provider / wire / history / lowering / model-turn / trajectory /
  compaction / session / reasoning suites: 2608 passed, 1 skipped, 0 failed.
- **Full local CI-filtered suite:** 14,268 passed, 64 skipped, 17 deselected, 0 failed, 0 errors (40:49).
- Behavioural differential against the reference implementation of the same transformation, executed
  as separate processes on different interpreters over 22 shared shapes: identical results on all 22
  (the pass fires on 9, leaves 13 untouched).
- Negative control: removing the call site fails 13 provider-level cases; restoring it passes all 30.


eous:
Thanks for the pull request and for digging into this. I agree there is a compatibility gap to address.

I tested the synthesized request from `170410b` directly against `gpt-6-astra` and `gpt-5.6-sol` with `store: false`. Both models:

- Reject the exact PR placeholder with **HTTP 400**, `array_above_max_length`: `input[1].content` must have a maximum length of zero.
- Reject the same placeholder with only `content` removed with **HTTP 404**: the fabricated `rs_roundrepair_...` item cannot be found under `store: false`.
- Accept the equivalent tool continuation with valid encrypted reasoning, or with the reasoning item omitted.

That confirms a commercial OpenAI regression when this repair runs. The behavior needs to be selected by an explicit endpoint/model capability rather than applied to every replay-enabled Responses request.

I also exercised actual compaction, persistence, and session reopen. The preserved tool-call turn retained its native reasoning, so I could not reproduce compaction stripping that turn's reasoning. Histories with missing reasoning or reasoning captured on another API surface remain relevant cases.

I'm working on the related issues:

- #1202 covers mandatory plaintext reasoning replay across the existing Chat Completions and Responses adapters, including the handling of missing history.
- #1203 investigates older stored OpenAI reasoning items without encrypted content. This remains an investigation: genuine, freshly returned bare reasoning IDs were accepted by both models in the live probes.

I appreciate you bringing this forward. I'm using these findings to work toward a capability-based fix that also preserves the commercial OpenAI path.

RedEyeNinja-BKK:
Thanks — your “Turnstone as an LLM ledger compiler” framing is useful, and it’s shaping the direction we’re taking outside this PR.

Rather than teach Switchyard to infer session semantics from Turnstone’s wire, we’re leaning toward a clean contract:

- Turnstone remains semantic authority for ledger/tool-round state, reasoning replay, compaction/resume, continuation, and provider/session quirks.
- Turnstone emits a small versioned execution intent describing hard requirements vs preferences, e.g. reasoning=forbidden, reasoning_replay=required, truthful_output_limit=required, min_context=N, local_preferred, escalation_allowed.
- Switchyard remains fleet/policy authority: match that intent against endpoint capabilities/readiness/cost/escalation, then lower simple endpoint controls such as Qwen enable_thinking=false.
- Switchyard should preserve backend semantic facts, not reconstruct the conversation ledger or synthesize missing reasoning/stop causes.

That seems consistent with your finding here that “replay enabled” is too coarse: actual endpoint/model replay behavior needs to be an explicit capability.

We’d migrate this shadow-first: emit/parse intent with no routing change, compare the hypothetical choice against today’s smart routes, then pilot one bounded local route.

Feedback welcome: does that boundary fit how you see Turnstone’s provider-neutral IR evolving? Is there already a natural place in Turnstone where semantic execution requirements/capabilities should live, rather than adding a parallel abstraction?

RedEyeNinja-BKK:
Taking your pointer — I'll implement the repair as a provider rather than as a flag on the OpenAI lane. Turn IR → lowering → provider makes it ours to own, and it keeps backend-specific repair out of `_openai_responses.py`.

On the measured rejections: both reproduce, and you were right that "replay enabled" is too coarse a selector. With `store: false` the synthesized item is rejected `400 array_above_max_length` when it carries `content`, and `404` when it does not, because the fabricated `rs_roundrepair_...` id has nothing to resolve against — while the same continuation is accepted with valid encrypted reasoning, or with the item omitted.

So this branch is reduced to the part that does not depend on that decision:

- `_ensure_responses_round_reasoning_items` is no longer selected by `replay_reasoning_to_model` alone. It now also needs a `synthesizes_round_reasoning` capability (`ModelCapabilities`, default `False`, settable from the model-definition capabilities JSON like the other operator-declared fields). With no declaration the item list is returned untouched, so the repair cannot reach an endpoint that has not opted in.
- Because the pass is inert by default, the three re-baselined wire goldens (`openai_responses__{native_orphan,native_reasoning,trailing_orphan}`) are restored to their upstream content: the default payload is byte-identical to before, and this branch changes the request sent to no endpoint that has not declared the capability.
- The battery declares the capability where it asserts the repair, and six cases pin the selection itself — capability off leaves the tool round untouched and byte-for-byte equal to a replay-off payload, replay off still wins, and the field is `False` by default.

Head is `f039a43a`. If the gate isn't worth carrying upstream, close it — nothing here blocks us, and the provider carries the real work. I'll follow #1202/#1203 and name the capability once rather than twice.

